PR #227, opened 2026-04-14
ci: pin all transitive Python deps with SHA-256 hashes

will-yuponce-db:
## Summary

- Expand `requirements.txt` files from top-level-only pins to full transitive lockfiles with SHA-256 hashes (generated via `uv pip compile --generate-hashes`)
- Add `--require-hashes` to CI `pip install` steps so pip rejects any package without a matching hash
- Update `scripts/lock-requirements.sh` to accept `UV_DEFAULT_INDEX` env var for internal PyPI mirrors (the URL is not written into lockfiles, keeping them portable for this public repo)

**Before:** 40 top-level pins, transitive deps resolved fresh each CI run
**After:** 126 + 112 + 14 fully pinned packages across `src/`, `src/backend/`, `src/e2e/` with hash verification

## Test plan

- [ ] CI pip install succeeds with `--require-hashes` against the new lockfiles
- [ ] Verify no internal proxy URLs leaked into committed files (`grep -r pypi-proxy .`)
- [ ] Run `./scripts/lock-requirements.sh` locally to confirm it regenerates cleanly

This pull request was AI-assisted by Isaac.


larsgeorge-db (approved):
lgtm

PR #312, opened 2026-04-30
fix(settings): use requester_email in role-request decision audit (#66)

### LINKED ISSUE #66: [Bug]: Role request decision endpoint throws 500 due to wrong field name
## Summary
Approving/denying a role access request via `POST /settings/roles/handle-request` returns 500 with an internal error toast.

## Repro
1. Submit a role access request.
2. As admin, handle request from Settings.
3. Observe failure toast: internal error during request decision.

## Root Cause
In `src/backend/src/routes/settings_routes.py`, audit logging references:
- `request_data.user_email`

But `HandleRoleRequest` model defines:
- `requester_email` (`src/backend/src/models/settings.py`)

This triggers an `AttributeError`, falls into the generic exception handler, and returns 500.

## Expected
Role request decision endpoint should complete successfully and return 200.

## Actual
500 error: "Failed to process role request decision due to an internal error."

## Suggested Fix
Replace `request_data.user_email` with `request_data.requester_email` in:
- `src/backend/src/routes/settings_routes.py`

Also add/adjust test coverage for this route to validate field usage.

SAY-5:
Fixes #66. `POST /settings/roles/handle-request` returned HTTP 500 because the audit log line referenced `request_data.user_email`, but the `HandleRoleRequest` model (`src/backend/src/models/settings.py:132`) declares the field as `requester_email`. The `AttributeError` propagated to the generic `except Exception` and surfaced to the user as the 'internal error' toast — confirmed reproducible by another reporter on the same issue.

## Change

`src/backend/src/routes/settings_routes.py:364` — `request_data.user_email` → `request_data.requester_email`. The audit log key (`'user_email'`) is preserved so existing audit consumers don't break.

## Test plan

- [x] AST parse on `settings_routes.py` is clean
- [x] Field name now matches `HandleRoleRequest` schema; the only previous reference path was the audit log line that triggered the 500

### DIFF
--- a/src/backend/src/routes/settings_routes.py
+++ b/src/backend/src/routes/settings_routes.py
@@ -361,7 +361,7 @@ async def handle_role_request_decision(
             feature='settings',
             action='HANDLE_ROLE_REQUEST',
             success=True,
-            details={'role_id': request_data.role_id, 'approved': request_data.approved, 'user_email': request_data.user_email}
+            details={'role_id': request_data.role_id, 'approved': request_data.approved, 'user_email': request_data.requester_email}
         )
         
         return result


larsgeorge-db (approved):
lgtm

larsgeorge-db:
Thank you!

PR #315, opened 2026-05-04
Feature/ontos approval workflows ux

mvkonchits-db:


src/backend/alembic/versions/g1_add_workflow_snapshot.py
```diff
@@ -0,0 +1,34 @@
+"""Add workflow_snapshot and workflow_name to agreement_wizard_sessions and agreements.
+
+Captures an immutable JSON snapshot of the workflow definition at session creation time,
+so historical agreements reflect what the signer actually saw (PRD #242, user stories #12-14, #23).
+
+Revision ID: g1_workflow_snapshot
+Revises: f1_merge_aa9_e2, 76a5bca4ce65
```
larsgeorge-db: ## The migration in this PR

`src/backend/alembic/versions/g1_add_workflow_snapshot.py`:

```python
revision: str = "g1_workflow_snapshot"
down_revision: Union[str, Sequence[str], None] = ("f1_merge_aa9_e2", "76a5bca4ce65")

def upgrade() -> None:
    op.execute("ALTER TABLE agreement_wizard_sessions ADD COLUMN IF NOT EXISTS workflow_snapshot TEXT")
    op.execute("ALTER TABLE agreement_wizard_sessions ADD COLUMN IF NOT EXISTS workflow_name VARCHAR(255)")
    op.execute("ALTER TABLE agreements ADD COLUMN IF NOT EXISTS workflow_snapshot TEXT")
    op.execute("ALTER TABLE agreements ADD COLUMN IF NOT EXISTS workflow_name VARCHAR(255)")

def downgrade() -> None:
    op.drop_column("agreements", "workflow_name")
    ...
```

It's seemingly intended as a merge-plus-schema-change revision. It then chains to two more migrations also in this PR: `h1_daimler_subscribe_on_behalf` → `h2_wizard_obo`.

## Bug: one of the two parents doesn't exist

The first parent `f1_merge_aa9_e2` exists and is itself a no-op merge of `aa9_is_approver` + `e2_semantic_display_name`. Confirmed.

The second parent `76a5bca4ce65` **does not exist anywhere in the repo**. I verified by:

1. Listing every file in `src/backend/alembic/versions/` at the PR's HEAD commit (`ce8de17`) and checking the `revision` line of each — none equals `76a5bca4ce65`.
2. Running a GitHub code search across the whole repo for the literal string — zero hits.
3. Diffing the alembic versions directory between `main` and the PR branch — only the 3 new files are added; nothing else changed there. So it's not a file that exists on main and was renamed in the PR.

The most likely candidate the author *meant* was the wizard-schema branch head `v4_wizard_completion_act` (the only obviously-related sibling head, since this migration touches `agreement_wizard_sessions` and `agreements` which are created by that revision). But that's a guess — the author should confirm.

## Practical impact

Running `alembic upgrade head` after this PR merges will fail with something like:

```
KeyError: '76a5bca4ce65'
```

…because Alembic can't resolve the parent. So this PR will break migrations on every fresh deploy and on any environment that needs to apply g1.

## Other things worth noting

- The `upgrade()` uses raw `ALTER TABLE ... ADD COLUMN IF NOT EXISTS` rather than `op.add_column`. That's Postgres-specific syntax. Fine if you're Postgres-only (your stack is Lakebase Postgres for prod and presumably local Postgres for dev), but it does make `downgrade()` asymmetric — `op.drop_column` doesn't have an `IF EXISTS` guard, so re-running downgrade will fail. Minor inconsistency.
- `workflow_snapshot` is `TEXT` rather than `JSONB`. For an "immutable JSON snapshot" on Postgres you'd typically want `JSONB` (queryable, indexable, validates JSON on insert). `TEXT` is the safer cross-db choice but loses those benefits — worth a comment in review.
- The fact that two new heads existed simultaneously (`f1_merge_aa9_e2` from one branch and the phantom `76a5bca4ce65` from somewhere else) suggests the author generated this with `alembic merge heads` against a local DB that had a revision the repo doesn't have — i.e., the migration was authored against a stale or hand-edited dev DB. That's the most plausible explanation for how a real-looking 12-char revision id ended up here without a corresponding file.

## Suggested review comment

> `g1_workflow_snapshot` lists `76a5bca4ce65` as a parent, but no revision file with that id exists in `src/backend/alembic/versions/` on either main or this branch. `alembic upgrade head` will fail with `KeyError: '76a5bca4ce65'`. Did you mean `v4_wizard_completion_act` (the wizard-schema head that introduced `agreement_wizard_sessions` / `agreements`)? Please regenerate this revision with `alembic merge heads` against a clean DB synced to this branch, and consider keeping the merge revision empty and putting the column additions in a follow-up child revision.

src/backend/src/common/config.py
```diff
@@ -159,6 +159,12 @@ class Settings(BaseSettings):
     UI_ABOUT_CONTENT: Optional[str] = Field(None, env='UI_ABOUT_CONTENT')  # Custom Markdown content for About page
     UI_CUSTOM_CSS: Optional[str] = Field(None, env='UI_CUSTOM_CSS')  # Custom CSS to inject into the app
 
+    # Welcome Disclaimer settings (first-open dialog, browser-flag-tracked)
+    # When enabled, shown to every user on app open until accepted; changing
+    # the text re-prompts everyone (config_version = hash of disclaimer text).
+    WELCOME_DISCLAIMER_ENABLED: bool = Field(False, env='WELCOME_DISCLAIMER_ENABLED')
+    WELCOME_DISCLAIMER_TEXT: Optional[str] = Field(None, env='WELCOME_DISCLAIMER_TEXT')
```
larsgeorge-db, line 166: Why are we not using an Approval Workflow with an "on_first_access" (or similar) trigger?

src/backend/src/common/workflow_executor.py
```diff
@@ -1915,11 +2228,73 @@ def resume_workflow(
...
+        # Cross-workflow variable propagation (#291):
+        # When the approval step was backed by a wizard session (approval flow),
+        # the user may have entered data (e.g. workspaceId, reason).  Merge those
+        # wizard step_results into context.step_results under the approval step's
+        # step_id so subsequent process-workflow steps can reference them via
+        # ${step_results.<approval_step_id>.<field>}.
+        if trigger_context and step_result:
+            try:
+                from src.repositories.agreement_wizard_sessions_repository import (
+                    agreement_wizard_sessions_repo,
```
larsgeorge-db, line 2247: Not a fan of using the repo directly. I would suggest only to use the appropriate managers/controllers (which can be injected) and then encode the business logic (here to load the user input) in a function of the manager.

src/backend/src/controller/data_products_manager.py
```diff
@@ -3021,6 +3072,153 @@ def subscribe(
...
+        except Exception as e:
+            # Non-fatal — never break the subscribe response on trigger errors.
+            logger.warning("on_subscribe trigger fire failed (non-fatal): %s", e)
+
+    # ------------------------------------------------------------------
+    # on_behalf_of validation (Daimler #486363)
+    # ------------------------------------------------------------------
+    # Tiny in-process cache to avoid hammering SCIM on each subscribe call
+    # within the same gunicorn worker. 60s TTL per spec.
+    _OBO_CACHE: Dict[str, tuple] = {}  # key -> (timestamp, exists_bool)
```
larsgeorge-db, line 3161: The workspace client is already wrapped into a CachingWorkspaceClient. Is that not enough to cache here as well?

src/backend/src/controller/settings_manager.py
```diff
@@ -272,6 +272,15 @@ def _load_persisted_settings(self) -> None:
                 self._settings.UI_CUSTOM_CSS = all_settings['UI_CUSTOM_CSS']
                 logger.info("Loaded UI_CUSTOM_CSS from database")
 
+            # Welcome disclaimer settings (first-open dialog)
+            if 'WELCOME_DISCLAIMER_ENABLED' in all_settings and all_settings['WELCOME_DISCLAIMER_ENABLED'] is not None:
+                self._settings.WELCOME_DISCLAIMER_ENABLED = all_settings['WELCOME_DISCLAIMER_ENABLED'].lower() == 'true'
+                logger.info(f"Loaded WELCOME_DISCLAIMER_ENABLED from database: {self._settings.WELCOME_DISCLAIMER_ENABLED}")
+
+            if 'WELCOME_DISCLAIMER_TEXT' in all_settings and all_settings['WELCOME_DISCLAIMER_TEXT'] is not None:
```
larsgeorge-db, line 280: See earlier comment. We should use an Approval Workflow.

src/backend/src/controller/settings_manager.py
```diff
@@ -272,6 +272,15 @@ def _load_persisted_settings(self) -> None:
...
+                self._settings.WELCOME_DISCLAIMER_TEXT = all_settings['WELCOME_DISCLAIMER_TEXT']
+                logger.info("Loaded WELCOME_DISCLAIMER_TEXT from database")
```
larsgeorge-db, line 282: This should not be INFO level logging. Maybe DEBUG at best.

src/backend/src/controller/settings_manager.py
```diff
@@ -1414,7 +1426,20 @@ def update_settings(self, settings: dict) -> Settings:
...
             self._settings.UI_CUSTOM_CSS = value
             logger.info(f"Updated UI_CUSTOM_CSS")
-        
+
+        # Welcome disclaimer settings (first-open dialog, browser-flag-tracked)
+        if 'welcome_disclaimer_enabled' in settings:
+            value = settings.get('welcome_disclaimer_enabled')
+            app_settings_repo.set(self._db, 'WELCOME_DISCLAIMER_ENABLED', str(value).lower() if value is not None else None)
+            self._settings.WELCOME_DISCLAIMER_ENABLED = bool(value) if value is not None else self._settings.WELCOME_DISCLAIMER_ENABLED
+            logger.info(f"Updated WELCOME_DISCLAIMER_ENABLED to: {value}")
```
larsgeorge-db, line 1435: See above

src/backend/src/controller/settings_manager.py
```diff
@@ -1414,7 +1426,20 @@ def update_settings(self, settings: dict) -> Settings:
...
+
+        if 'welcome_disclaimer_text' in settings:
+            value = settings.get('welcome_disclaimer_text') or None
+            app_settings_repo.set(self._db, 'WELCOME_DISCLAIMER_TEXT', value)
+            self._settings.WELCOME_DISCLAIMER_TEXT = value
+            logger.info(f"Updated WELCOME_DISCLAIMER_TEXT")
```
larsgeorge-db, line 1441: See above

src/backend/src/data/default_workflows.yaml
```diff
@@ -1032,25 +1033,83 @@ workflows:
...
     steps:
-      - id: "reason"
-        name: "Why are you subscribing?"
+      - id: "review-terms"
+        name: "Review Terms"
+        type: "legal_document"
+        config:
+          title: "Terms of Service"
+          description: "Please review the following terms before subscribing."
+          body_markdown: "# Data Product Terms of Service\n\nBy subscribing to this data product, you agree to the following:\n\n1. **Acceptable Use** — You will use the data only for authorized business purposes.\n2. **Confidentiality** — You will not share raw data outside your organization.\n3. **Retention** — Data will be retained according to your organization's data retention policy.\n4. **Compliance** — You will comply with all applicable data protection regulations.\n5. **Termination** — The data product owner may revoke access for policy violations."
```
larsgeorge-db, line 1042: We need to consider i18n

src/backend/src/data/demo_data_migrate.sql
```diff
@@ -0,0 +1,56 @@
+-- ============================================================================
+-- Schema Migration: COMPREHENSIVE — all missing columns and tables
```
larsgeorge-db, line 2: What is this file for?

src/backend/src/routes/approvals_routes.py
```diff
@@ -138,6 +181,178 @@ async def abort_approval_session(
...
+    request: Request,
+    db: DBSessionDep,
+    entity_type: Optional[str] = Query(None),
+    entity_id: Optional[str] = Query(None),
+    limit: int = Query(50, ge=1, le=100),
+    _: bool = Depends(PermissionChecker('settings', FeatureAccessLevel.READ_ONLY)),
+):
+    """List agreements, optionally filtered by entity type/id."""
+    from src.repositories.agreements_repository import AgreementsRepository
+    repo = AgreementsRepository()
```
larsgeorge-db, line 197: As a general rule: Routes should NEVER call the repositories or databases directly, but use a manager/controller function.

src/backend/src/routes/approvals_routes.py
```diff
@@ -138,6 +181,178 @@ async def abort_approval_session(
...
+
+
+@router.get('/approvals/agreements/{agreement_id}')
+async def get_agreement(
+    agreement_id: str,
+    db: DBSessionDep,
+    _: bool = Depends(PermissionChecker('settings', FeatureAccessLevel.READ_ONLY)),
+):
+    """Get a single agreement by ID."""
+    from src.repositories.agreements_repository import AgreementsRepository
```
larsgeorge-db, line 209: See above

src/backend/src/routes/approvals_routes.py
```diff
@@ -138,6 +181,178 @@ async def abort_approval_session(
...
+    _: bool = Depends(PermissionChecker('settings', FeatureAccessLevel.READ_ONLY)),
+):
+    """Download the agreement as a PDF document.
+
+    Generation priority:
+    1. Pre-generated PDF file on disk (if ``pdf_storage_path`` exists).
+    2. On-the-fly PDF via fpdf2 (``build_agreement_pdf``).
+    3. Fallback to styled HTML if fpdf2 is not installed.
+    """
+    from src.repositories.agreements_repository import AgreementsRepository
```
larsgeorge-db, line 248: Same

src/backend/src/routes/approvals_routes.py
```diff
@@ -138,6 +181,178 @@ async def abort_approval_session(
...
+    if agreement.step_results:
+        try:
+            parsed = json.loads(agreement.step_results) if isinstance(agreement.step_results, str) else agreement.step_results
+            if isinstance(parsed, list):
+                step_results = parsed
+        except (json.JSONDecodeError, TypeError):
+            pass
+
+    # Try real PDF generation via fpdf2
+    if _HAS_FPDF:
```
larsgeorge-db, line 318: All of this business logic should be in the controller.

src/backend/src/tests/unit/test_workflow_executor_templates.py
```diff
@@ -204,6 +205,88 @@ def test_none_entity_name_resolves_empty(self):
         assert substitute_template("N=${entity_name}", ctx) == "N="
 
 
+# =========================================================================
+# Daimler go-live: list/dict serialization + nested context paths
```
larsgeorge-db: As noted elsewhere, applies globally

src/frontend/src/components/common/welcome-disclaimer-dialog.tsx
larsgeorge-db, line 1: See earlier notes on rather using the Approval Workflow feature with a new trigger. That way users can configure what they want instead of one block of text.

larsgeorge-db (changes requested):

PR #324, opened 2026-05-05
fix: commit singleton session mutations to prevent data loss on restart

larsgeorge-db:
## Summary

- **Bug**: All singleton managers (`SettingsManager`, `ConnectionsManager`, `DataProductsManager`, `SemanticLinksManager`) use a long-lived `self._db` session created at startup. Their mutation methods only called `flush()` — which sends SQL to the DB within the transaction — but never `commit()`. On app restart, PostgreSQL rolls back the uncommitted transaction, silently reverting all API-made changes (role group assignments, connection configs, product status transitions, etc.).
- **Fix**: Replace `flush()` with `commit()` in all singleton-session mutation paths across 4 manager files.
- **Impact**: Fixes data loss observed in Marketplace deployments where admin role groups reverted to defaults after every restart.

## Changed Files

| File | Changes |
|------|---------|
| `settings_manager.py` | `create_app_role`, `update_app_role`, `delete_app_role`, list backfill |
| `connections_manager.py` | `create_connection`, `update_connection`, `delete_connection`, `ensure_system_connection`, `migrate_from_app_settings` |
| `data_products_manager.py` | `transition_status`, `publish_product` |
| `semantic_links_manager.py` | `create_link` |

## Test plan

- [ ] Update admin role groups via Settings UI, restart app, verify groups persist
- [ ] Create/update/delete a connection, restart app, verify changes persist
- [ ] Transition a data product status, restart app, verify status persists
- [ ] Verify no regressions in request-scoped session paths (normal CRUD via API)

### DIFF
--- a/src/backend/src/controller/connections_manager.py
+++ b/src/backend/src/controller/connections_manager.py
@@ -91,7 +91,7 @@ class ConnectionsManager:
             created_by=created_by or "",
         )
         self._db.add(db_obj)
-        self._db.flush()
+        self._db.commit()
         self._db.refresh(db_obj)
         logger.info(f"Created connection '{data.name}' (type={data.connector_type})")
         return ConnectionResponse.model_validate(db_obj)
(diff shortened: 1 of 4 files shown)


mvkonchits-db (approved):
lgtm

PR #289, opened 2026-04-23
fix(schema-importer): stop recursive schema listing under table nodes

larsgeorge-db:
## Summary

Fixes the Schema Importer bug where expanding a table node re-listed every sibling table from the parent schema (recursive listing).

- **Connector contract:** `DatabricksConnector.list_assets` now returns `[]` for paths with 3+ segments (leaf FQN: `catalog.schema.object[.column...]`), matching the documented BigQuery connector contract. Previously, anything with `len(parts) >= 2` fell into the schema-listing branch using `parts[0]` / `parts[1]`, causing the recursion.
- **Browse enrichment:** `SchemaImportManager.browse` now sources columns from `connector.get_asset_metadata(path).schema_info.columns` and appends them as `column` browse nodes when expanding a leaf asset. Failures degrade silently to listing-only nodes (per PRD failure model). Root browses (`path is None or ""`) skip metadata to stay cheap.
- **No frontend, schema, or persistence changes.** `BrowseNode` shape is unchanged (already supports `node_type`, `path`, `has_children`, `description`).

Closes #287
Refs PRD #286 (`docs/prds/prd-schema-importer-table-expand-recursion.md`)
Plan: `plans/schema-importer-table-expand-recursion.md`

## Changes

| File | Change |
| --- | --- |
| `src/backend/src/connectors/databricks.py` | Add `len(parts) >= 3 → return []` guard in `list_assets`; update docstring |
| `src/backend/src/controller/schema_import_manager.py` | Append column nodes from `get_asset_metadata().schema_info` after listing; dedupe; silent-degrade on errors |
| `src/backend/tests/test_schema_importer_browse.py` | New: 10 unit tests (path-depth contract + column enrichment) |
| `docs/prds/prd-schema-importer-table-expand-recursion.md` | PRD |
| `plans/schema-importer-table-expand-recursion.md` | Implementation plan |
| `docs/notes/work-item-schema-importer-table-expand-recursion.md` | Initial scoping note |

## Test plan

- [x] Unit: \`hatch -e dev run pytest src/backend/tests/test_schema_importer_browse.py\` (10 passed)
- [ ] Manual UI: Schema Importer → Databricks connection → expand catalog → schema → table; confirm columns render (with descriptions where available) and **no** sibling tables appear under the table node
- [ ] Manual UI: Re-collapse / re-expand the table; tree state remains consistent
[...]

PR #325, opened 2026-05-05
fix(schema-importer): expand functions/MVs/streaming tables; preserve source-system labels

larsgeorge-db:
## Summary

Follow-up to #289. Two coupled gaps in the Schema Importer were preventing users from drilling into the columns level for non-table leaf assets.

### Bug 1: `has_children=false` on column-bearing leaves

`_has_children` only included `UC_TABLE`, `UC_VIEW`, `BQ_TABLE`, `BQ_VIEW`, `SNOWFLAKE_TABLE`, `SNOWFLAKE_VIEW` — so the UI never rendered an expand chevron on:

- UC: materialized views, streaming tables, **functions** (parameters)
- BigQuery: materialized views, external tables, routines
- Snowflake: materialized views, functions, procedures

All of these populate `schema_info` via `get_asset_metadata()`, so they should be expandable. Models, volumes, and datasets stay non-expandable on purpose — they have no column-level schema.

### Bug 2: `node_type` collapsed to "system" / "ml model"

`_display_type` went through `_TYPE_MAP`, which intentionally collapses source-system concepts to a smaller Ontos asset-type vocabulary used at *import* time (e.g. `UC_FUNCTION` / `UC_VOLUME` -> `"System"`). At *browse* time we want the source label so users can tell a function from a volume, and so the frontend picks the right icon.

Introduces `_DISPLAY_TYPE_MAP` that preserves: `function`, `procedure`, `model`, `volume`, `view`, `table`, `routine`, `dataset`, `project`, `database`.

### Frontend

`nodeIconMap` extended:
- `function` / `procedure` -> Braces (same as routine — code-like)
- `volume` -> HardDrive

## Files

| File | Change |
| --- | --- |
| `src/backend/src/controller/schema_import_manager.py` | New `_DISPLAY_TYPE_MAP`; rewrite `_display_type`; extend `_has_children` |
| `src/backend/tests/test_schema_importer_browse.py` | +3 tests: expandable types, non-expandable types, display labels |
| `src/frontend/src/components/schema-importer/schema-browser.tsx` | Add `function` / `procedure` / `volume` icons |

## Test plan

- [x] Unit: `hatch -e dev run pytest backend/tests/test_schema_importer_browse.py` (13/13 passed)
- [x] Playwright E2E: navigated to `system.ai.python_exec` (UC function); chevron rendered, expanding showed `code` parameter as a column node, row label reads `function` (not `system`); `bge_*` models read `model` (not `ml model`)
- [ ] Regression: `samples.tpch.customer` still expands cleanly to its 8 columns (verified in #289, no logic change to the table path)
- [ ] Manual: BigQuery / Snowflake connections (no live test envs in this run; covered by code review of the connector contracts)

## Out of scope

- Backend cache invalidation when display labels change (current dev reload + page refresh works fine)
- Persisting column selections in the import payload

### DIFF
--- a/src/backend/src/controller/schema_import_manager.py
+++ b/src/backend/src/controller/schema_import_manager.py
@@ -866,31 +866,97 @@ class SchemaImportManager:
 # Module-level helpers
 # ---------------------------------------------------------------------------
 
+# Per-asset-type display labels for the browse tree (drives the
+# `node_type` field on BrowseNode and the icon picked by the frontend).
+#
+# Distinct from `_TYPE_MAP` on purpose: `_TYPE_MAP` collapses source-system
+# concepts to a smaller Ontos asset-type vocabulary used at *import* time
+# (e.g. UC_FUNCTION/UC_VOLUME -> "System"), but at *browse* time we want
+# to preserve the source label so users can tell a function from a volume.
+_DISPLAY_TYPE_MAP: Dict[str, str] = {
+    # Databricks / Unity Catalog
+    UnifiedAssetType.UC_CATALOG.value: "catalog",
+    UnifiedAssetType.UC_SCHEMA.value: "schema",
+    UnifiedAssetType.UC_TABLE.value: "table",
+    UnifiedAssetType.UC_VIEW.value: "view",
+    UnifiedAssetType.UC_MATERIALIZED_VIEW.value: "view",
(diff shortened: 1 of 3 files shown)


mvkonchits-db (approved):
lgtm

PR #331, opened 2026-05-05
feat(demo-data): split demo data into 5 standalone vertical presets

larsgeorge-db:
## Summary

Replace the implicit "retail base + optional industry overlay" model with five fully self-contained demo packs, one per vertical, loaded via a single `preset` parameter.

- Renamed `demo_data.sql` → `demo_data_retail.sql` (default preset).
- API now takes a single preset param: `POST /api/settings/demo-data/load?preset=<retail|hls|fsi|mfg|auto>`. No more implicit "base + overlay" layering.
- Each vertical pack (`hls`, `fsi`, `mfg`, `auto`) is standalone — own data domains, business roles, delivery methods, vertical asset types, tag namespaces+tags, RDF concepts, semantic links, workflows, assets, lineage, owners, ratings, compliance runs+results, cost items, subscriptions.
- Hardened `DELETE /api/settings/demo-data`:
  - Each DELETE runs in its own savepoint so legacy-table absence cannot poison the outer transaction.
  - Real failures now surface in the response under a `skipped` array (status becomes `partial`); legacy-table absences are demoted to debug log.
  - Added missing UUID prefix patterns: `0f1%`/`0f3%`/`0f5%` (assets), `0f6%`/`0fb%` (`business_owners`), and demo-only filtered deletes for `business_roles`, `delivery_methods`, vertical `asset_types` via `created_by='system@demo'`.
- Fixed retail RDF collection-metadata block: switched from `ON CONFLICT (id)` to bare `ON CONFLICT DO NOTHING` so logical-key collisions with app-bootstrapped `urn:meta:sources` collection metadata no longer abort the INSERT batch.

## Why

Loading `hls`/`fsi`/`mfg`/`auto` previously required retail to be loaded first (or at minimum produced confusing mixed data, e.g. "POS Transaction Stream" appearing after loading HLS on an empty DB). Verticals diverge meaningfully on foundational/reference data, so a shared retail base was the wrong abstraction.

## API change

| Before | After |
| --- | --- |
| `POST /api/settings/demo-data/load` (implicit retail) and optional industry overlays | `POST /api/settings/demo-data/load?preset=<name>` — exactly one self-contained file |

The `preset` parameter defaults to `retail`, so existing scripts continue to work.

## Verification

- Cycle `clear → load → clear` for all 5 presets via REST: every preset loads standalone with **bleed=0** and clears with **skipped=0**.
- 3× repeat retail load+clear is idempotent (907 records cleared each cycle).
- Backend log inspected during the cycle — **no warnings or errors**.

| Preset | Stmts | Domains | Products | Assets |
| --- | --- | --- | --- | --- |
| retail | 63 | 13 | 7 | 78 |
| hls | 39 | 6 (+2 shared) | 5 | 8 |
| fsi | 39 | 6 (+1 shared) | 5 | 8 |
| mfg | 39 | 6 (+2 shared) | 5 | 8 |
| auto | 39 | 6 (+1 shared) | 5 | 8 |

## Test plan

- [ ] Pull branch, restart backend, hit `POST /api/settings/demo-data/load?preset=retail` then `DELETE /api/settings/demo-data` → confirm clean wipe (no `skipped` entries in response).
- [ ] Repeat for `hls`, `fsi`, `mfg`, `auto` on an empty DB — each should load standalone (no retail bleed in the UI).
- [ ] Inspect Asset Explorer / Data Products / Glossary / Workflows / Tags for each preset — content should be vertical-flavored.
- [ ] Tail `/tmp/backend.log` during load+clear cycles — confirm no `Statement execution warning` lines.

## Notes

- Vertical packs are intentionally lighter than retail (8 assets / 5 products vs. 78 / 7) — they're "one of each kind" demos, expandable later.
- Documentation updated: `docs/Ontos Setup.md` (preset table), `STRUCTURE.md`, banner comments in `app.py` / `compliance_routes.py`, `startup_tasks.py` docstrings, and all 5 SQL file headers.

### DIFF
--- a/.planning/codebase/STRUCTURE.md
+++ b/.planning/codebase/STRUCTURE.md
@@ -116,7 +116,9 @@ ucapp/
 - Purpose: Store seed data, demo fixtures, ontology definitions
 - Contains:
   - `taxonomies/` — RDF ontologies (e.g., `ontos-ontology.ttl`)
-  - Demo data files (loaded via POST /api/settings/demo-data/load)
+  - Demo data files: one self-contained pack per preset, named `demo_data_{preset}.sql`
+    (presets: `retail` (default), `hls`, `fsi`, `mfg`, `auto`).
+    Loaded via `POST /api/settings/demo-data/load?preset=<name>`.
 
 **`src/backend/src/tests/`** (Test Suite)
 - Purpose: Unit and integration tests
--- a/docs/Ontos Setup.md
+++ b/docs/Ontos Setup.md
@@ -283,7 +283,26 @@ Once the app is up, click on the app's URL to open it.
 
 ## Step #5: (Optionally) Load Demo Data
 
-Ontos ships with demo data, which can be loaded on demand. Navigate to the `/docs` page in Ontos (by editing the URL in the browser's address bar):
+Ontos ships with five **standalone demo packs** — one per industry — and the
+`POST /api/settings/demo-data/load` endpoint accepts a `preset` query
+parameter to pick exactly one of them:
+
+| `preset`  | File                       | Vertical flavour                                |
+| --------- | -------------------------- | ----------------------------------------------- |
+| `retail`  | `demo_data_retail.sql`     | Retail / e‑commerce (default)                   |
+| `hls`     | `demo_data_hls.sql`        | Healthcare & Life Sciences (clinical, claims)   |
+| `fsi`     | `demo_data_fsi.sql`        | Financial Services (banking, capital markets)   |
+| `mfg`     | `demo_data_mfg.sql`        | Manufacturing (production, quality, EHS)        |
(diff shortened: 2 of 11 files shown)


mvkonchits-db (approved):
Not critical (created by Isaac):
 
 - ⚠️  ?industry=hls callers will silently get retail (param removed). I checked — only caller is src/frontend/src/tests/helpers/demo-data.ts which sends no params, so it still works. No UI surface uses
  ?industry=.... Safe.                                                                                     
  - ⚠️  Legacy-table detection uses "does not exist" in str(e) substring match — fragile to driver/locale 
  changes but cosmetic, the DELETE still rolls back cleanly.                                               
  - ⚠️  One spot in the load path references nested in the except without guarding against begin_nested()
  itself failing — practically unreachable, but inconsistent with the more defensive pattern in clear.

PR #341, opened 2026-05-06
chore(settings): drop unused exception bindings

will-yuponce-db:
## Summary

SonarQube (rule `python:S1481`) flagged 13 `except Exception as e:` sites in `settings_routes.py` where `e` is never referenced — each block already logs via `logger.error(..., exc_info=True)`, which captures the full traceback on its own. Dropped the unused binding.

No behavior change. Pure cleanup.

Sites: L45, L105, L128, L150, L162, L172, L182, L241, L742, L773, L798, L828, L849.

## Test plan

- [ ] CI: existing backend tests pass (no logic touched)
- [ ] Manual: trigger one of the affected endpoints with a forced failure and confirm the traceback still shows up in logs

### DIFF
--- a/src/backend/src/routes/settings_routes.py
+++ b/src/backend/src/routes/settings_routes.py
@@ -42,7 +42,7 @@ async def get_settings_route(manager: SettingsManager = Depends(get_settings_man
     try:
         settings_data = manager.get_settings() # Renamed variable to avoid conflict
         return settings_data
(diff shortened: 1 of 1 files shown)


will-yuponce-db:
Dropping — these were MINOR severity. Scope reduced to CRITICAL findings; #343 covers the remaining piece.

PR #318, opened 2026-05-05
Feature/ontos approval workflows ux

mvkonchits-db:
Summary
                                                                                                                                                            
  Closes PRD #242 (Approval Workflows v2) — 24 of 25 user stories shipped — plus the v1 go-live feature set (subscribe-on-behalf, consumer groups, in-wizard
   principal selection, on_first_access disclaimer trigger) and the supporting infrastructure to make all of it work end-to-end against UC Volumes.         
                                                                                                                                                          
  ▎ Note for reviewers: this PR replaces #315 (which was opened from a fork — fork-PR CI couldn't get the JFrog OIDC token at the                           
  ▎ databrickslabs-protected-runner-group stage). Same diff, same author, full review thread from #315 is migrated below in the "Lars's PR #315 review —  
  ▎ addressed" section. CI runs in upstream context here so checks should green.                                                                            
                                                                                                                                                          
  PRD #242 — Approval Workflows v2 user-story matrix                                                                                                        
   
  Visual step types                                                                                                                                         
                                                                                                                                                          
  - #1 Designer toolbox shows approval-specific step types                                                                                                  
  - #2 Visual steps share title + markdown description
  - #3 legal_document step (markdown body, scroll-to-end, ack checkbox)                                                                                     
  - #4 acknowledgement_checklist step (items with required flags + 10-item cap)                                                                             
  - #5 co_signers step (min/max, principal_type)                                                                                                            
                                                                                                                                                            
  Non-visual step types                                                                                                                                     
                                                                                                                                                            
  - #6 persist_agreement non-visual step                                                                                                                    
  - #7 generate_pdf with configurable storage destination (UC Volume via SDK Files API)
  - #8 deliver step with channels + recipients (email deferred — see #16)                                                                                   
                                                                                                                                                            
  Wizard UX                                                                                                                                                 
                                                                                                                                                            
  - #9 Wizard shows clear title + informational paragraph                                                                                                   
  - #10 Non-visual steps show "Processing…" tile and auto-advance
  - #11 Wizard refuses to advance on validation failure                                                                                                     
  - #12 Wizard remains stable when admin edits workflow mid-session (immutable snapshot)                                                                    
  - #22 Wizard heading labels current step's purpose                                                                                                        
                                                                                                                                                            
  Audit & PDF                                                                                                                                               
                                                                                                                                                            
  - #13 Agreement includes immutable workflow snapshot                                                                                                      
  - #14 PDF generated from snapshot, not live workflow
[...]

PR #345, opened 2026-05-07
fix(data-products): null-safety in change-impact analyzer for products without team/description/support

mvkonchits-db:
## Summary

Found while debugging a customer 500 on `PUT /api/data-products/{id}` (adding an output port). **Not a regression from #318** — latent null-safety bug exposed by a permitted-but-undertested state: a data product with `owner_team_id IS NULL` and no row in `data_product_teams`. The response builder serializes that as `team: None`, and the change-impact analyzer crashes on `old_product.get('team', {}).get('members', [])`.

### The Python footgun

`dict.get('k', default)` only falls back to `default` when the key is **missing**. When the key is **present but the value is None**, the default is skipped and the chained `.get(...)` raises `AttributeError: 'NoneType' object has no attribute 'get'`. Confirmed in DB inspection of the affected deployment with team-less products — schema permits it (`owner_team_id` is nullable), the API returns the product fine, only the analyzer trips.

## Call sites fixed

In `src/backend/src/utils/product_change_analyzer.py`:

- `analyze()`: `team`, `description`, `support`, `inputPorts`, `outputPorts`, `managementPorts` — switched every `x.get('k', default)` (where the value can legitimately be `None`) to `x.get('k') or default`. The two-level `team → members` access uses `(x.get('team') or {}).get('members') or []`.
- `_analyze_port_changes()`: filter out `None` entries when building the port name → port map.
- `_analyze_team_changes()`: same defensive filter for member list, plus `or ''` on the name/email key fallback so a member with both fields `None` doesn't blow up the dict comprehension.
- `_analyze_support_changes()`: same defensive filter for support channel list.
- `role.lower()` calls now use `role = member.get('role') or 'unknown'` so a stored `role=None` doesn't crash.

## Regression test

New `src/backend/src/tests/unit/test_product_change_analyzer.py` covers:

1. All top-level optional fields = `None` (smoke test — analyzer returns a clean empty-change result, no raise).
2. `team: None` → populated team (analyzer recognizes a feature-level addition).
3. `team` present but `members: None` (second-level None path).
4. `outputPorts: None` → populated ports (feature-level).
5. `support: None` on both sides.
6. `description: None` on both sides.
7. Team member with `role=None` (covers `role.lower()` crash path).

All 7 cases pass; existing tests untouched.

## Ship plan

Small, isolated, no schema or API changes. Pure null-safety hardening in one utility module + a new unit test file.

### DIFF
--- a/src/backend/src/tests/unit/test_product_change_analyzer.py
+++ b/src/backend/src/tests/unit/test_product_change_analyzer.py
@@ -1,0 +1,148 @@
+"""
+Unit tests for ProductChangeAnalyzer null-safety.
+
+Regression coverage for a customer 500 where a data product with no team
+(`owner_team_id IS NULL`, no `data_product_teams` row) crashed the analyzer
+on `old_product.get('team', {}).get('members', [])`. The chained-`.get()`
+default only fires for missing keys; when the key is present but the value
+is None (the common API serialization case), the default is skipped and
+`.get(...)` blows up on NoneType.
+"""
+from typing import Dict
+
+import pytest
+
+from src.utils.product_change_analyzer import (
+    ChangeType,
+    ProductChangeAnalyzer,
+)
+
+
+def _empty_product(version: str = "1.0.0") -> Dict:
+    """Product with all top-level optional collections explicitly set to None.
+
+    Mirrors the response-builder output for a data product whose team,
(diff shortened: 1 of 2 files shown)


larsgeorge-db (approved):
lgtm

PR #348, opened 2026-05-07
fix(data-products): default request-type to 'access' via configurable prop

mvkonchits-db:
## Summary

P1 fix from a customer feedback batch. `RequestProductActionDialog` defaulted `requestType` to `'status_change'` — a rarely-used path. The most common entry point (access requests) had to flip the dropdown manually on every open.

## Approach

Rather than hardcoding `'access'` as the new default and risking regressions in publish/certify entry points, this PR adds a configurable `defaultRequestType` prop with a sensible global default of `'access'`. Call sites can override when their context calls for a different starting type.

## Changes

- **`request-product-action-dialog.tsx`** — new `defaultRequestType?: RequestType` prop, defaults to `'access'`. Wired through the `useState` initializer and the open-reset `useEffect` (added to deps array).
- **`request-product-action-dialog.test.tsx`** (new) — three regression cases:
  - No prop → renders the access-request title
  - `defaultRequestType="status_change"` → renders `Request Status Change`
  - `defaultRequestType="status_change"` + `canDirectStatusChange` → renders `Change Status`

## Call sites

Only one consumer in the codebase: `data-product-details.tsx:2154`, a generic "Request..." button with no status-change-specific intent. Left without an explicit prop so it picks up the new `'access'` default — which matches the user expectation in the customer feedback.

## Verification

- Frontend tests: 400 pass / 6 pre-existing skips (matches main)
- Typecheck: clean (`yarn tsc --noEmit`)
- No backend / API / schema changes

Small, isolated, low-risk.

### DIFF
--- a/src/frontend/src/components/data-products/request-product-action-dialog.test.tsx
+++ b/src/frontend/src/components/data-products/request-product-action-dialog.test.tsx
@@ -1,0 +1,93 @@
+/**
+ * Tests for RequestProductActionDialog component
+ *
+ * Focused on the initial-request-type behavior. The dialog uses Radix Select
+ * elements which hang in jsdom when interacted with, so these tests only assert
+ * static text rendered by the dialog header — which is driven directly by the
+ * initial request-type state.
+ */
+import { screen } from '@testing-library/react';
+import { describe, it, expect, vi, beforeEach } from 'vitest';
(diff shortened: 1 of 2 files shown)


larsgeorge-db (approved):
lgtm

PR #350, opened 2026-05-08
fix(data-products): restrict Link Asset picker to ontology-supported deliverable types

mvkonchits-db:
## Summary

Reported by a customer: linking a Catalog (or Schema) asset to a Data Product's output port returned **422** from `/api/entity-relationships`:

```
ValueError: Ontology does not allow relationship 'portHasTable' from 'OutputPort' to 'Catalog'
```

Confirmed scope decision with the team: the ontology intentionally models port-to-asset relationships only for *deliverable* asset types (Table, View, Dataset, APIEndpoint, MLModel) — **not** container types like Catalog or Schema. Catalogs and Schemas are organisational containers, not units a port "delivers." This PR makes the asset picker UX match that intent.

## Why this is a UX fix, not a security fix

The backend validator at `entity_relationships_manager.py:70-98` already returns 422 for any invalid (predicate, range) tuple. That's the security layer and it's correct as-is. This PR is purely about not surfacing options in the picker that would always fail at submit time — and not silently constructing nonsense predicates like `portHasCatalog` from a string template.

## Changes

- **`src/frontend/src/views/data-product-details.tsx`**
  - New exported `PORT_TO_ASSET_PREDICATE` map (5 entries — mirrors `ontos-ontology.ttl:838-891`)
  - `handleLinkAssets` now looks up the predicate via the map. If a caller somehow gets an unsupported asset through, the API call is skipped with a `console.warn` rather than constructing an invalid predicate string
  - `<AssetSelector>` invocation drops the misleading hardcoded `relationshipType="portHasTable"` (it was being stamped onto every selection regardless of asset type) and passes `targetAssetTypes={PORT_DELIVERABLE_ASSET_TYPES}` instead, so the picker filters its display to the 5 supported types
  - Description text updated to: *"Only deliverable asset types (Table, View, Dataset, API Endpoint, ML Model) can be linked to an output port."*

- **`src/frontend/src/views/data-product-details.test.tsx`** (new) — 10 tests covering the predicate map and selection logic for each supported type plus skip-on-unsupported behaviour
- **`src/frontend/src/components/common/asset-selector.test.tsx`** (new) — 3 component tests confirming Catalog/Schema rows are hidden when `targetAssetTypes` is set, backward compat preserved when omitted, and helper text rendered

## Verification

- Frontend tests: 413 pass / 6 pre-existing skips
- `yarn build` succeeds
- Backend validation unchanged — still returns 422 for any unsupported (predicate, range) tuple, confirming the security gate is intact

## Out of scope

- **Ontology extensions:** if/when port-to-Catalog or port-to-Schema relationships become a real product feature, the change would be (a) add new predicates to `ontos-ontology.ttl`, (b) add the new types to `PORT_TO_ASSET_PREDICATE`. Backend validator inherits the new tuples automatically. Not in this PR.
- Backend authorization model — unchanged.

### DIFF
--- a/src/frontend/src/views/data-product-details.tsx
+++ b/src/frontend/src/views/data-product-details.tsx
@@ -69,6 +69,30 @@ import { ApprovalEntity } from '@/types/settings';
 type ViewMode = 'minimal' | 'medium' | 'large'
 const VIEW_MODE_STORAGE_KEY = 'data-product-view-mode'
 
+/**
+ * Output port → asset relationship predicates.
+ *
+ * Mirrors the `portHas*` predicates declared in `ontos-ontology.ttl` (lines
+ * 838–891). The ontology intentionally restricts these relationships to
+ * *deliverable* asset types — not container types like Catalog or Schema.
+ * The picker filter and the POST predicate selection both flow from this map,
+ * so they stay in lock-step with the ontology.
+ *
+ * If a new deliverable type is added to the TTL, add it here. If a caller
+ * tries to link an unsupported type the POST is skipped with a warning;
+ * the backend at `entity_relationships_manager.py` is the actual security
+ * gate and will return 422 for any invalid (predicate, range) tuple.
+ */
+export const PORT_TO_ASSET_PREDICATE: Readonly<Record<string, string>> = {
+  Table: 'portHasTable',
+  View: 'portHasView',
+  Dataset: 'portHasDataset',
+  APIEndpoint: 'portHasEndpoint',
+  MLModel: 'portHasModel',
+};
+
+const PORT_DELIVERABLE_ASSET_TYPES = Object.keys(PORT_TO_ASSET_PREDICATE);
+
 type CheckApiResponseFn = <T>(
   response: { data?: T | { detail?: string }, error?: string | null | undefined },
   name: string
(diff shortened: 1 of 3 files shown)


src/frontend/src/views/data-product-details.tsx
```diff
@@ -69,6 +69,220 @@ import { ApprovalEntity } from '@/types/settings';
...
+  if (lowerStatus === 'retired' || lowerStatus === 'deprecated') return 'outline';
+  return 'default';
+}
+
+/**
+ * Statuses where a Data Product can still be edited in-place. Anything with a
+ * higher lifecycle status (`active`, `retired`, `deprecated`) must be cloned
+ * for editing instead.
+ */
+export const IN_PLACE_EDITABLE_STATUSES: ReadonlyArray<string> = [
```
larsgeorge-db, line 159: This should not be hardcoded into the UI. This business logic should be handled in the controller/manager. The UI should call an endpoint to get that information or it should be part of the endpoint already used when loading the entity itself.

src/frontend/src/views/data-product-details.tsx
```diff
@@ -69,6 +69,220 @@ import { ApprovalEntity } from '@/types/settings';
...
+/**
+ * Pure ViewMode → section visibility predicate. The component-level wrapper
+ * just closes over the current `viewMode`.
+ */
+export function shouldShowSectionForViewMode(viewMode: ViewMode, section: string): boolean {
+  switch (viewMode) {
+    case 'minimal':
+      return ['deliverables', 'description', 'hierarchy'].includes(section);
+    case 'medium':
+      return !['management-ports', 'support-channels', 'metadata-panel', 'ratings', 'costs', 'quality'].includes(section);
```
larsgeorge-db, line 117: A note for later: This should not be hardcoded in the view, as this may be configurable by users/admin eventually. Need to create a ticket for it.

larsgeorge-db (approved):
lgtm

PR #351, opened 2026-05-08
fix(frontend+backend): wire ConsumerGroupsPicker, expose on_first_access trigger, fix Select overflow, resolve real SCIM groups in /api/user/details (PR #318 follow-up)

mvkonchits-db:
## Summary

Fixes three frontend wiring gaps left over from PR #318, a UX overflow bug in the workflow designer's trigger dropdown, **and** a backend bug where `/api/user/details` returned the user's email as the only "group" instead of their real workspace SCIM group memberships. All five are small, additive, well-tested.

### 1. Wire `ConsumerGroupsPicker` into the live edit/create dialog

PR #318 added the picker component but mounted it in `data-product-form-dialog.tsx`, which is **not imported anywhere** in the app — the active dialog is `data-product-create-dialog.tsx`. As a result, no user could ever populate `consumer_principals` from the UI, even though the backend column, validators, repo, and workflow template variable (`${entity.consumer_principals}`) were all in place.

This PR mounts the picker in the active dialog. Backend untouched.

### 2. Expose `on_first_access` in the workflow designer's trigger dropdown

The `on_first_access` trigger type was added to the backend `TriggerType` enum in #318 (`process_workflows.py:70`) but never made it into the frontend `TriggerType` union. Same for `EntityType.USER`. Result: users could not pick the trigger or entity type when creating an `on_first_access` workflow in the designer — even though the workflow type works fine when created via API.

Adds:
- `'on_first_access'` to the FE `TriggerType` union
- `'user'` to the FE `EntityType` union
- Both to `ALL_TRIGGER_TYPES` / `ALL_ENTITY_TYPES`
- `on_first_access: ['user']` to `SUPPORTED_TRIGGER_ENTITY_MAP`
- i18n labels in `en/common.json`

### 3. Fix `<SelectContent>` overflow with auto-scroll

The trigger type dropdown in the Workflow Designer overflows the viewport when more than ~20 items are listed. Root cause: the shadcn/ui `Viewport` was pinning its height to the trigger's height, blocking Radix's auto-scroll. Removed that misplaced style and added `max-h-[var(--radix-select-content-available-height)] overflow-y-auto` to `SelectContent`.

### 4. Resolve real SCIM groups in `/api/user/details`

**Symptom:** `GET /api/user/details` returned `{"groups": ["<user_email>"]}` even when the user actually belonged to multiple workspace groups (`admins`, `users`, etc.). This broke the subscribe-on-behalf wizard's "For a group I'm part of" picker — the only option shown was the user's own email.

**Root cause:** `UsersManager.get_current_user` (OBO path) and `get_user_details_by_email` (SP fallback path) both unconditionally appended the user's email as an "implicit group" alongside whatever SCIM returned, *and* `current_user.me()` / `users.list` often return `groups=[]` depending on token scope and requested attributes — so the email-as-implicit-group fallback became the only entry every time.

**Fix:**
- New `UsersManager._resolve_scim_groups()` helper. Takes the SCIM user from `me()` / `users.list` and, if `.groups` is empty, re-fetches the full record via `users.get(id)`. The OBO client can read its own SCIM membership without admin permissions.
- `get_user_details_by_email` now passes `attributes="id,userName,displayName,emails,groups,active"` to `users.list` so the SCIM list endpoint actually returns group memberships in workspaces where the SP has SCIM read access.
- When real groups are resolved, the email is **not** appended — the picker / permissions layers now see actual workspace groups.
- Email-as-implicit-group fallback is preserved for FEVM-style workspaces where SCIM read isn't available, so role rules that match on email continue to work.

**Verified end-to-end** against the deployed app: with OBO header set, `GET /api/user/details` now returns `{"groups": ["admins","users","account-ops-users"]}` (previously: `["<user_email>"]`).

## Verification

- Frontend tests: 487 → 494 pass / 6 pre-existing skips
- Backend unit tests: 705 passed, 1 skipped (full suite); `test_users_manager.py` 19/19 pass (4 pre-existing tests updated, 5 new tests added covering OBO real-groups path, `users.get(id)` fallback path, and email-as-implicit-group fallback)
- New FE tests: 7 in `data-product-create-dialog.test.tsx` + 7 in `workflow-labels.test.ts`
- `yarn tsc --noEmit` clean
- Coverage gate: not regressed
- E2E: `/api/user/details` responses captured before / after fix on the deployed app

## Out of scope

- Deletion of the dead `data-product-form-dialog.tsx` — separate cleanup PR
- Frontend changes for the `/api/user/details` response — none needed; the picker already consumes `groups` correctly

## Files changed

```
src/backend/src/controller/users_manager.py                                    (+ helper, attributes param, fallback gating)
src/backend/src/tests/unit/test_users_manager.py                               (+ updated 4 tests, 5 new tests)
src/frontend/src/components/data-products/data-product-create-dialog.tsx       (+31 -0)
src/frontend/src/components/data-products/data-product-create-dialog.test.tsx  (+170, new)
src/frontend/src/components/ui/select.tsx                                       (+2 -2)
src/frontend/src/i18n/locales/en/common.json                                    (+2 -2)
src/frontend/src/lib/workflow-labels.test.ts                                    (+45, new)
src/frontend/src/lib/workflow-labels.ts                                         (+5 -0)
src/frontend/src/types/process-workflow.ts                                      (+5 -0)
```

### DIFF
--- a/src/backend/src/common/authorization.py
+++ b/src/backend/src/common/authorization.py
@@ -115,8 +115,20 @@ async def get_user_details_from_sdk(
 
     # Try using OBO client with current_user.me() first (no admin permissions required)
     obo_token = request.headers.get('x-forwarded-access-token')
+    # Diagnostic: surface which forwarded headers actually arrive from the
+    # Databricks Apps proxy. Without these at INFO level we can't tell the
+    # difference between "OBO header missing" vs "OBO header present but empty"
+    # vs "scope not granted" when triaging a deployed app.
+    fwd_email = request.headers.get('X-Forwarded-Email')
+    fwd_user = request.headers.get('X-Forwarded-User')
+    logger.info(
+        "auth.headers: x-forwarded-access-token=%s, X-Forwarded-Email=%s, X-Forwarded-User=%s",
+        "present" if obo_token else "MISSING",
+        fwd_email or "<none>",
+        fwd_user or "<none>",
+    )
     if obo_token:
-        logger.debug("Using OBO token with current_user.me() for user lookup (no admin permissions required).")
+        logger.info("Using OBO token with current_user.me() for user lookup (no admin permissions required).")
         try:
             obo_client = get_obo_workspace_client(request, settings)
             user_info_response = manager.get_current_user(obo_client=obo_client, real_ip=real_ip)
@@ -132,7 +144,10 @@ async def get_user_details_from_sdk(
             raise HTTPException(status_code=500, detail="An unexpected error occurred")
 
     # Fallback: Use get_user_details_by_email if no OBO token (requires admin permissions)
-    logger.debug("No OBO token available, falling back to get_user_details_by_email (requires admin permissions).")
+    logger.info(
+        "No OBO token (x-forwarded-access-token) on request, falling back to SP-scoped get_user_details_by_email. "
+        "If groups appear missing in the response, the app's user_authorization scopes likely need iam.current-user:read."
+    )
     user_email = request.headers.get("X-Forwarded-Email")
     if not user_email:
         user_email = request.headers.get("X-Forwarded-User")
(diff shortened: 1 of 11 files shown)


larsgeorge-db (approved):
lgtm

PR #352, opened 2026-05-08
fix(data-products): wire ConsumerGroupsPicker into the live edit/create dialog

mvkonchits-db:
## Summary

Fixes a wiring gap from #318. The `ConsumerGroupsPicker` was mounted in `data-product-form-dialog.tsx`, which is **not imported anywhere** — the active dialog opened from `data-product-details.tsx` and `data-products.tsx` is `data-product-create-dialog.tsx`. As a result, users had no way to populate `consumer_principals` from the UI even though the backend (column, Pydantic models, repository, `${entity.consumer_principals}` workflow template variable) was already in place.

This PR mounts the picker in the active dialog so the round-trip works end-to-end.

## Changes

- `data-product-create-dialog.tsx`
  - Add `consumer_principals` to the form schema and default values
  - Hydrate from `product.consumer_principals` in edit mode
  - Include in both `POST /api/data-products` (create) and `PUT /api/data-products/{id}` (edit) payloads
  - Add a **Consumer Groups** section between **Tags** and **Save Changes** with the same description copy as the dead form-dialog (typed-principal explainer + `${entity.consumer_principals}` reference)
- `data-product-create-dialog.test.tsx` (new)
  - Picker renders when dialog opens in edit mode
  - Initial values from the product prop hydrate as chips
  - Submit payload includes `consumer_principals` after a free-text add
- No backend changes required.

## Test results

- `yarn tsc --noEmit` — clean
- `yarn test --run --coverage` — **403 passed | 6 skipped (409)**, function coverage **75.92%** (gate: 25%)
- New file adds 3 component tests, all green

## E2E verification (deployed app)

Verified against a deployment:

1. Synced changed frontend files to the workspace and ran `databricks apps deploy` — succeeded, app `RUNNING`.
2. `GET /api/data-products/{id}` — `consumer_principals: []` (clean draft).
3. `PUT /api/data-products/{id}` with body containing `consumer_principals: [{type: "group", value: "<group-name>"}]` — accepted, response echoes the principal.
4. `GET /api/data-products/{id}` again — value persisted.
5. `GET` after re-fetch shows the same shape that the picker emits — confirms the wire format dialog → API → DB → API → dialog matches end-to-end.

The deployed `data-product-create-dialog.tsx` in the workspace shows 11 references to `ConsumerGroupsPicker` / `consumer_principals` (vs 0 on `main`), confirming the synced code is the live code path.

## Out of scope

Deletion of the dead `data-product-form-dialog.tsx` (whose own test file is `describe.skip`) is intentionally not included here to keep the diff focused on the wiring fix. Will be addressed in a follow-up.

### DIFF
--- a/src/frontend/src/components/data-products/data-product-create-dialog.test.tsx
+++ b/src/frontend/src/components/data-products/data-product-create-dialog.test.tsx
@@ -1,0 +1,170 @@
+/**
+ * Tests for DataProductCreateDialog — focused on the Consumer Groups picker
+ * being mounted in the *active* edit/create dialog. Earlier work mounted the
+ * picker only in `data-product-form-dialog.tsx`, which is dead code (not
+ * imported anywhere). This dialog is the one actually opened from the UI, so
+ * the picker has to live here for users to populate `consumer_principals`.
+ */
+import { screen, fireEvent, waitFor } from '@testing-library/react';
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { renderWithProviders } from '@/test/utils';
+import DataProductCreateDialog from './data-product-create-dialog';
+import type { DataProduct } from '@/types/data-product';
+
+// Mock external service hooks so the dialog mounts cleanly.
+vi.mock('@/hooks/use-toast', () => ({
+  useToast: () => ({ toast: vi.fn() }),
+}));
+
+vi.mock('@/hooks/use-domains', () => ({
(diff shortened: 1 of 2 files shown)


mvkonchits-db:
Superseded by #351 — that PR bundles the Consumer Groups picker wiring with two related PR #318 follow-up fixes (`on_first_access` trigger type + Select overflow). Closing this one to keep review focused on the bundle. Thanks!

PR #353, opened 2026-05-08
feat(workflows): wire approval wizard for all for_request_* triggers (Path B)

mvkonchits-db:
## Summary

Closes the FE wiring gap from PR #318 — the approval wizard infrastructure was generic but only invoked from the Subscribe path. This PR introduces a portable launch helper used by all `for_request_*` actions on a Data Product, enabling a deployment-specific configuration to ship custom-field wizards on Request Access, Request Review, Request Publish, Request Certify, and Request Status Change without code changes per workflow.

## What changes

**Frontend**

- New `useApprovalWizardTrigger()` hook with `lookupWorkflowId(triggerType)` (`src/frontend/src/hooks/use-approval-wizard-trigger.ts`). Replaces three hardcoded `for_subscribe` lookups in `data-product-details.tsx`, `marketplace-view.tsx`, and `discovery-section.tsx`.
- `RequestProductActionDialog`: before submit, looks up the corresponding `for_request_<type>` workflow. If configured, opens `ApprovalWizardDialog` with `autoStartWithPreselected`; on completion the collected wizard fields are merged into `wizard_data` on the original API submit body. If no workflow is configured, the dialog falls through to today's direct-submit behavior. Direct status changes (`canDirectStatusChange=true`) skip the wizard entirely.
- `ApprovalWizardDialog.onComplete` extended with an optional third `wizardFields` argument that aggregates every `user_action` step's payload across the session. Backward compatible — existing callers (Subscribe flows) ignore the extra arg.

**Backend**

- `AccessGrantRequestCreate` gets `wizard_data: Optional[Dict[str, Any]]` plus `model_config = ConfigDict(extra="allow")` so the body surface is flexible for any deployment-specific fields the workflow author defines without per-customer model edits.
- `AccessGrantsManager.create_request` splats `wizard_data` into `entity_data` at the top level so Process workflow steps can reference values via `${entity.<field_id>}`. The full `wizard_data` bag is also preserved for authors who prefer `${entity.wizard_data.<id>}`. First-class fields take precedence on collision (workflow authors should namespace custom field ids). Top-level extra fields surfaced by `extra="allow"` also forward.
- `APP_ACTION_TRIGGER_TYPES` allowlist on `GET /api/workflows/for-trigger/{trigger_type}` gains `FOR_REQUEST_CERTIFY` — was defined in `TriggerType` but not exposed via the route, blocking the certify wizard.

## Tests

**FE — 7 new tests**, all passing alongside the existing 407:

- `use-approval-wizard-trigger.test.ts` (4): success, null-on-no-data, null-on-throw, URL passthrough for arbitrary trigger types.
- `request-product-action-dialog.test.tsx` (3): direct submit when no workflow is configured (no `wizard_data` on body); wizard launch with merged fields on completion; direct status change skips the wizard lookup.

**BE — 6 new tests**, all 711 unit tests pass:

- `test_access_grant_wizard_data.py`: model accepts `wizard_data` and arbitrary extras; manager splats wizard fields into `entity_data`; first-class fields are not clobbered by colliding wizard ids; legacy shape preserved when no wizard runs; top-level extras also forward.

## Test plan

- [ ] Reviewer runs `cd src/frontend && yarn test --run` — expect 414 passed.
- [ ] Reviewer runs `cd src/backend && hatch -e dev run pytest src/tests/unit/ -q` — expect 711 passed.
- [ ] Live verification — configure a `for_request_access` workflow with a `user_action` step that has 2-3 custom `required_fields` on a deployed Ontos app. Click Request Access on a Data Product. Verify the wizard opens, the fields render, the request submits with the field values, and `on_request_access` Process workflows fire with the field values in their `entity_data`.

## Out of scope (intentionally deferred)

- **Workspace derivation helper** (DP → output ports → catalog/schema → workspace metadata): not required for Path B's success and adds a non-trivial chain across managers; deferred to a follow-up.
- **Refactor of Subscribe's hardcoded launch to use the new hook** (cosmetic — Subscribe was already using inline lookups; the three Subscribe call sites are now updated to use the new hook in this PR, but no broader Subscribe refactor was attempted).
- **OBO step inside request-action wizards** (orthogonal — already supported by the wizard's `on_behalf_of` step type).
- **No new manifest scopes** (already in PR #351).
- **No changes to existing Subscribe completion behavior** (only added the generic launch alongside).

## Live verification status

The author's demo workspace is currently in a stuck deploy state from a previous round, so live E2E was not run. Coverage relies on the unit-test fakes that trace the full path from `wizard_data` on the request body through to the trigger's `entity_data` argument. Live verification should be picked up by the next deploy cycle.

### DIFF
--- a/src/backend/src/controller/access_grants_manager.py
+++ b/src/backend/src/controller/access_grants_manager.py
@@ -144,21 +144,47 @@ class AccessGrantsManager:
         
         db.commit()
         
-        # Fire the ON_REQUEST_ACCESS trigger
+        # Fire the ON_REQUEST_ACCESS trigger.
+        # Path-B portable wizard launch: the FE merges fields collected by the
+        # ``for_request_access`` approval wizard's ``user_action`` steps into
+        # ``data.wizard_data``. We splat that dict at the top level of
+        # ``entity_data`` so Process workflow steps can reference each field
+        # via ``${entity.<field_id>}`` without the workflow author needing to
+        # know the wrapper key. First-class fields take precedence on
+        # collision (workflow authors should namespace custom field ids).
+        # Pydantic ``extra='allow'`` may also bag other ad-hoc fields onto the
+        # model — we forward those too to keep the surface flexible.
+        entity_data: Dict[str, object] = {
+            "request_id": str(request_db.id),
+            "entity_type": data.entity_type,
+            "entity_id": data.entity_id,
+            "entity_name": data.entity_name,
+            "requested_duration_days": data.requested_duration_days,
+            "permission_level": data.permission_level.value,
+            "reason": data.reason,
+        }
+        wizard_data = getattr(data, "wizard_data", None) or {}
+        if isinstance(wizard_data, dict) and wizard_data:
+            for k, v in wizard_data.items():
+                if k not in entity_data:
+                    entity_data[k] = v
+            # Preserve the original namespaced bag too so workflow authors who
+            # prefer ``${entity.wizard_data.<id>}`` over the splatted form can
+            # use either. Only set when non-empty so legacy callers without a
+            # wizard see the original ``entity_data`` shape unchanged.
+            entity_data["wizard_data"] = wizard_data
+        # Forward any other ``extra='allow'`` fields from the request body.
+        extra_fields = getattr(data, "model_extra", None) or {}
+        for k, v in extra_fields.items():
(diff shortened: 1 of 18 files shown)


larsgeorge-db (dismissed):
lgtm

larsgeorge-db (approved):

PR #376, opened 2026-05-19
fix(workflows): per-trigger permission dispatch + workflow_version migration + wizard fetch-by-id

mvkonchits-db:

---

## Summary

Five related fixes to the workflow + permissions surface that surfaced when extending the access-grant wizard from administrator-only to end-user use:

1. **`feat(alembic)`** — adds the missing `workflow_version` column on the `agreements` table. The column was already on the SQLAlchemy model but no migration existed, so any greenfield deployment would silently drift from the model.
2. **`fix(workflows)`** — replaces the static feature-permission gates on wizard endpoints with a per-trigger dispatch table (`WIZARD_PERMISSION_DISPATCH`). Each workflow trigger now declares the right feature/level pair instead of forcing the same `process-workflows:READ_WRITE` gate on every persona.
3. **`fix(settings)`** — tolerates stale `feature_id` entries in role-permission JSON. If a role config references a feature that no longer exists in `features.py`, the role merge skips that entry rather than raising; prevents legacy role configs from bricking the settings UI.
4. **`fix(perms)`** — drops two read-only feature gates on `business-owners` endpoints that were causing the wizard to 403 for users with `process-workflows:READ_WRITE` but not `business-owners:READ_WRITE`. The business-owners write is now driven by workflow-trigger context.
5. **`fix(workflows)`** — wizard now fetches a single workflow by id (with trigger-aware permission dispatch) instead of relying on `list-workflows`, which the user may not have permission for.

## Why these are bundled

All five surfaced together while shipping the access-grant wizard from admins to Data Consumers. They share the same root cause: the wizard's endpoint surface was built when only admins used it, so every endpoint assumed admin-level permissions and admin-only side-effects. Extending the persona without auditing the permission scope created consistent 403s on the same trigger path.

## Safety contracts

- **Migration**: `i1_workflow_version` is additive only; nullable column with no backfill. No backwards-compatibility risk.
- **Permission dispatch**: defaults preserve prior behavior for existing triggers — only new triggers (access-grant wizard from end users) gain new entries.
- **Verification**: customers with custom role configs should audit `feature_permissions` against the new dispatch table after merge. Roles previously requiring `business-owners:READ_WRITE` but not the per-trigger feature may need updating.

## Tests

- `test_workflow_version_migration.py` — migration applies and column exists
- `test_wizard_permission_dispatch.py` — each trigger → feature+level mapping
- `test_settings_manager.py` — stale-feature tolerance
- Manual UI verification on FEVM and a customer test workspace clone; access-grant wizard launches for both administrators and Data Consumers

## Notes for reviewers

The PR also includes unrelated CI plumbing changes (removal of internal JFrog-only actions; promotion of `test-coverage.yml`) that came along on the same branch from `main` drift. Happy to split these out if preferred.

### DIFF
--- a/src/backend/alembic/versions/i1_workflow_version.py
+++ b/src/backend/alembic/versions/i1_workflow_version.py
@@ -1,0 +1,28 @@
+"""Add workflow_version column to agreements table
+
+Closes the model/migration gap: AgreementDb.workflow_version was added to
+the SQLAlchemy model in PRD #242 work but no Alembic migration was
+generated. Fresh deployments hit a 500 in the wizard's _complete_session
+because the INSERT references this missing column.
+
+Revision ID: i1_workflow_version
+Revises: h3_rename_consumer_principals
+Create Date: 2026-05-18
+"""
+from alembic import op
+
+
+revision = "i1_workflow_version"
+down_revision = "h3_rename_consumer_principals"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.execute(
+        "ALTER TABLE agreements ADD COLUMN IF NOT EXISTS workflow_version INTEGER"
+    )
+
+
+def downgrade() -> None:
+    op.execute("ALTER TABLE agreements DROP COLUMN IF EXISTS workflow_version")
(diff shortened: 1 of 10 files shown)


larsgeorge-db (approved):
lgtm

PR #377, opened 2026-05-19
feat(workflows): business-owner resolver traversal for access_grant proxy entities

mvkonchits-db:

---

## Summary

When a workflow's `entity_type` is an "indirection" entity that names another underlying entity by id — currently `access_grant` (which references a data product or output port) — the business-owner resolver should look up business-owner assignments on the *underlying* entity, not on the proxy.

Before this PR, `BusinessOwnersManager.get_business_owners_for_entity` filtered strictly on `BusinessOwnerDb.object_id == context.entity_id`. For an access-grant workflow execution, `context.entity_id` was the `AccessGrantDb.id` — so no business owner ever matched even when the grant's underlying data product had a perfectly valid BO assignment. Approvals routed nowhere.

This PR introduces a `PROXY_ENTITY_TYPES` registry (currently `{'access_grant': resolver_fn}`) and updates the resolver to traverse from the proxy entity to its underlying entity before looking up business-owner assignments.

## Safety contracts

- **Stacked on `#376`** — depends on PR #376's per-trigger permission dispatch (the workflow now reaches this resolver via the new dispatch). Merge order: #376 first.
- **Scope is conservative**: only `access_grant` is wired into `PROXY_ENTITY_TYPES`. Non-proxy entities use the unchanged direct lookup path. No risk of routing surprises for `data_product`, `data_contract`, or other direct-entity workflows.
- **Verification**: customers with existing access-grant approval workflows will start routing to the underlying DP's business owners. This is the intended behavior — the previous "routes nowhere" state was the bug. Recommend customers verify their BO assignments before depending on this routing post-merge.

## Tests

- `test_pr_b_resolver.py` — 11/11 passing on FEVM (covers proxy-entity traversal, direct-entity preservation, no-BO-found fall-through, edge cases for missing access grants)
- Manual UI verification on a a customer test workspace clone: notification routed to the correct BO of the underlying DP; workflow completed end-to-end

### DIFF
--- a/src/backend/src/common/workflow_executor.py
+++ b/src/backend/src/common/workflow_executor.py
@@ -269,12 +269,44 @@ def _resolve_role_to_users(
             logger.warning(f"Business role {br_id} not found")
             return []
 
+        # Some trigger contexts carry a "proxy" entity that stands in for the
+        # real entity that Business Owners are assigned to. For example:
+        #   - on_request_access fires on entity_type=access_grant, where
+        #     context.entity_id is the access-grant request's UUID, NOT the
+        #     underlying data product. Business Owners are assigned to data
+        #     products, so a naive lookup by context.entity_id returns [].
+        # For these proxy entity types, the trigger firer puts the underlying
+        # entity's id under context.entity['entity_id']. We traverse to that
+        # so the BusinessOwnerDb lookup hits the right object.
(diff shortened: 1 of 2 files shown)


larsgeorge-db (dismissed):
lgtm

PR #378, opened 2026-05-19
feat(settings): role switcher — membership-scoped + backend hardening

mvkonchits-db:

---

## Summary

Adds a role switcher to the settings UI that lets a user with multiple assigned roles act under one chosen role for the session. Useful for testing other personas (e.g., a Data Producer testing the Data Consumer experience) without needing to log in as a different user.

Backend hardening:
- Membership computation now considers email-as-group (a single user can be in a role's `assigned_groups` by email directly, not just via a workspace group)
- The applied-role override is session-scoped via `SettingsManager.get_applied_role_override_for_user` and consumed by `enforce_feature_permission` so the user sees the chosen role's effective permissions across all endpoints
- `teams_manager.get_teams_for_user` filters team membership through the applied role, so a role-switched user's team list reflects what the chosen role would see (not the union of all their actual memberships)

## Safety contracts

- **Independent**: no PR dependencies. Off `main`.
- **Switcher only appears for users who have ≥ 2 assigned roles**. Single-role users see no UI change.
- **Cannot escalate**: a user can only switch to roles they're already assigned to. The switcher does not grant new permissions.
- **Falls back gracefully**: if `applied_role_id` resolution fails, the user sees their permanent permissions unchanged.

## Tests

- Manual UI verification on FEVM: a Mikhail account with admin stripped (representing a multi-role non-admin user) sees the 2-radio switcher, swap to Data Consumer correctly shrinks visible features, swap back restores them.

## Known limitations

- Role switching cannot fully simulate a Data Producer / Data Consumer experience for a user who is *also* a workspace admin — the workspace-admin status (checked via `is_user_admin(user_groups, settings)`) still short-circuits many cascade-bypass decisions, so the user will continue to see admin-only resources even when role-switched. Documented as a known interaction; recommended workaround is to use a separate non-admin user for thorough persona simulation.

### DIFF
--- a/src/backend/src/controller/settings_manager.py
+++ b/src/backend/src/controller/settings_manager.py
@@ -279,11 +279,29 @@ class SettingsManager:
             logger.warning(f"Failed to load persisted settings from database: {e}")
 
     # --- Role override helpers (in-memory persistence) ---
-    def set_applied_role_override_for_user(self, user_email: Optional[str], role_id: Optional[str]) -> None:
+    def set_applied_role_override_for_user(
+        self,
+        user_email: Optional[str],
+        role_id: Optional[str],
+        caller_groups: Optional[List[str]] = None,
+        caller_is_admin: bool = False,
+    ) -> None:
         """Sets or clears the applied role override for a user.
 
         When role_id is None, the override is cleared and the user's actual group-based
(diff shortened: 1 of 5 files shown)


larsgeorge-db (approved):
lgtm

Issue #379, opened 2026-05-19
[PRD]: Tracking: workflow + permission + data-product hardening — 2026-05 series


mvkonchits-db:
Tracking issue for a coordinated set of pull requests delivering wizard-permission scoping, business-owner routing fixes, data-product visibility cascade, partial-update safety, and clone-for-editing schema-drift repair.

These changes are filed and reviewed as a group rather than individually to preserve the safety contracts between PRs (see below).

### Wizard + workflow surface (filed)
- [x] #368 — feat(workflows): in-UI editor for user_action required_fields
- [x] #369 — feat(workflows): trigger-picker UX overhaul
- [x] #376 — fix(workflows): per-trigger permission dispatch + workflow_version migration + wizard fetch-by-id  *(PR A)*  ✅ verified
- [x] #377 — feat(workflows): business-owner resolver traversal for access_grant proxy entities  *(PR B, stacked on #376)*  ✅ verified
- [x] #378 — feat(settings): role switcher — membership-scoped + backend hardening  *(PR C)*  ✅ verified

### Data-product surface (to be filed once verified on a production-style workspace)
- [x] #395 — feat(data-products): ownership-cascade scoping + clone schema-drift repair  *(PR D + Bug S; stacked on #376)*  ✅ verified
- [x] #396 — feat(data-products): delivery method preservation + project_id preservation on create  *(PR E + Bug Q)*  ✅ verified
- [x] #397 — fix(data-products): UUID serializer + missing schema fields  *(PR H + Bug R)*  ✅ verified
- [x] #398 — fix(data-products): exclude_unset on PUT  *(PR I)*  ✅ verified
- [x] #393 — fix(workflows): per-execution authz on handle-approval; gate -> notifications:READ_WRITE  *(PR K, stacked on #376)*  ✅ verified
- [x] #394 — fix(workflows): relax handle-approval outer gate to notifications:READ_ONLY  *(PR L, stacked on #393)*  ✅ verified
- [x] #399 — fix(data-products): admin cascade-bypass via auth manager  *(PR O; stacked on #395)*  ✅ verified
- [x] #380 — feat(access-grants): enrich entity_data with underlying-DP fields at trigger time  *(PR M; lets webhook templates surface underlying DP consumer_principals)*  ✅ verified

## Dependency graph

```
#376 (A)
  ├── #377 (B) ─ stacked
  └── PR D ─ stacked (when filed)
        └── PR O ─ stacked (when filed)

[...]

PR #380, opened 2026-05-19
feat(access-grants): enrich entity_data with underlying-DP fields at trigger time

mvkonchits-db:

---

## Summary

When an access-grant request fires its `ON_REQUEST_ACCESS` trigger, the workflow's trigger context carries `entity_type=access_grant` and `entity_id=<request_id>` — the AGR is a proxy for the underlying entity (today: `data_product`). Webhook `body_template` authors typically need fields from that underlying entity. Specifically:

- **`consumer_principals`** carries the AD/UC access group(s) that downstream provisioners (e.g. an Azure Automation runbook) need to add the requester into for catalog access.
- **`data_product_name`** is useful for log-friendly webhook bodies.

Before this change, those fields were unavailable to the template — `${entity.consumer_principals}` resolved to nothing because `entity_data` only carried AGR fields, never the underlying DP's. Workarounds were brittle: hard-code the group in the template, modify the FE to inject `consumer_principals` into the AGR POST body, or wait for a fetch-underlying-entity step type that doesn't exist in the engine.

## What it does

After caller-supplied keys (`wizard_data` / `extra='allow'` overrides) are merged into `entity_data`, the manager now:

1. Checks if `data.entity_type == "data_product"`
2. Fetches the underlying DP via the repository
3. Deserializes the DP's `consumer_principals` JSON-string column into an array
4. Adds `consumer_principals` and `data_product_name` to `entity_data` — but only if the caller didn't already supply them (caller wins)

Failures (DP not found, malformed JSON, transient DB error) are logged but never propagate. The enrichment is a nice-to-have; access-grant submission must continue to work even if the enrichment can't run.

## Safety contracts

- **Independent**: no PR dependencies. Off `main`.
- **Caller-supplied keys are preserved**: `wizard_data.consumer_principals` and `extra='allow'.consumer_principals` keep working unchanged.
- **Narrow scope**: only `data_product` underlying entities are enriched today. Other types fall through unchanged. Adding more is one-line.
- **Fail-safe**: a missing DP or unreadable column is logged but never breaks the AGR creation flow.
- **No DB migration**: pure read of an existing column at request time.

## Verification

Customers running access-grant workflows where the webhook needs the data product's access group can now reference `${entity.consumer_principals}` directly in `body_template` — no FE change, no workflow-config gymnastics. The change is also a safety net for workflows that previously templated this field expecting it to work (they were getting an empty string today).

## Tests

`test_access_grant_dp_enrichment.py` — 6/6 pass:
- Happy path: DP carries `consumer_principals` → appears on `entity_data`
- DP with no consumers → empty array, not `None`
- Caller-supplied keys (`extra='allow'`) win over enrichment
- Non-`data_product` entity types are NOT enriched (no DP fetch happens)
- DP-fetch failures are fail-safe (request still goes through)
- Malformed JSON in `consumer_principals` is logged and degrades to `[]`

### DIFF
--- a/src/backend/src/controller/access_grants_manager.py
+++ b/src/backend/src/controller/access_grants_manager.py
@@ -8,6 +8,7 @@ Manages time-limited access grants to assets including:
 - Triggering expiry workflows
 """
 
+import json
 import uuid
 from datetime import datetime, timedelta, timezone
 from typing import Dict, List, Optional, Tuple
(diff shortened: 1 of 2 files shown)


larsgeorge-db (approved):
lgtm

PR #393, opened 2026-05-20
fix(workflows): per-execution authz on handle-approval; outer gate -> notifications:READ_WRITE

mvkonchits-db:

---

## Summary

Adds per-execution authorization on `POST /api/workflows/handle-approval` and corrects the outer feature gate. Before this change, the route had only an outer feature gate (`settings:READ_WRITE`) and any user clearing that gate could approve any execution by guessing the `execution_id` — a horizontal privilege escalation risk.

This PR:
1. Adds `_assert_caller_authorized_for_execution()` — verifies the caller is the direct recipient (by email/username), a member of the recipient role's `assigned_groups`, or an admin (intentional bypass for rescue/manual intervention).
2. Changes the outer gate from `settings:READ_WRITE` (semantically wrong — approving a workflow notification isn't a settings action) to `notifications:READ_WRITE`. PR L (#TBD) relaxes this further to `READ_ONLY`; the two PRs together arrive at the design-correct gate.
3. Refactors `WIZARD_PERMISSION_DISPATCH` to declare the same gate for `for_approval_response` wizard launches.

## Safety contracts

- **Stacked on #376** (PR A). The dispatch table extension uses PR A's `WIZARD_PERMISSION_DISPATCH` infrastructure.
- **Must be merged before #TBD** (PR L). PR L relaxes the outer gate further; the per-execution check introduced here is what makes that safe.
- **No customer-impact** for admin users — they retain access via the admin bypass in `can_user_access_notification`. Non-admin Business Owners would still be blocked by the outer `:READ_WRITE` gate; PR L solves that.

## Tests

`test_handle_approval_per_execution_check.py` — 7 tests covering:
- Dispatch table pins `("notifications", READ_WRITE)` (will be flipped to `READ_ONLY` by PR L)
- `_find_approval_notifications_for_execution` returns only rows for the given execution
- `_assert_caller_authorized_for_execution` raises 403 when no notification exists
- 403 when caller is not recipient and no role match
- 200 when caller is the direct recipient
- 200 via recipient role membership (group overlap)
- Short-circuits on first match (no over-fetch)

Verified end-to-end on FEVM: BO can approve their own AGR (3/3 iterations, ~1.4s median); non-BO is blocked by inner check (admin-bypass intentional — see brief in tracking issue).

### DIFF
--- a/src/backend/src/routes/workflows_routes.py
+++ b/src/backend/src/routes/workflows_routes.py
@@ -11,7 +11,7 @@ from fastapi import APIRouter, Depends, HTTPException, Request, Query
 from sqlalchemy.orm import Session
 
 from src.common.database import get_db
-from src.common.dependencies import DBSessionDep, AuditManagerDep, AuditCurrentUserDep
+from src.common.dependencies import DBSessionDep, AuditManagerDep, AuditCurrentUserDep, get_notifications_manager
 from src.common.authorization import (
     PermissionChecker,
     enforce_feature_permission,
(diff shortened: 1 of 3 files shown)


larsgeorge-db (dismissed):
lgtm

PR #395, opened 2026-05-20
feat(data-products): ownership-cascade scoping + clone-product schema-drift repair

mvkonchits-db:

Stacked on #376 (PR A).

## 🇬🇧 Plain English

**Two things, related to data products:**

1. **Before this PR**, the *Products* page showed every data product to every non-admin user — even DPs they had nothing to do with. A Data Producer on the sales team could see the marketing team's DPs and the finance team's DPs side by side. This PR scopes the list to **what the caller actually has a stake in**: data products in their project, owned by their team, or that they created.

2. **Also fixed in this PR**: clicking "Edit" on an already-published data product was completely broken — the backend tried to clone the DP first (the standard versioned-editing pattern) but the clone code referenced columns that no longer exist on the database models (`expectation`, `dataset_name`, `sboms`, `spdx_*`, plus a handful more). The clone code would throw `AttributeError` on the first output port, or `IntegrityError` on the first management port / support channel / team member. So every "Edit on an active DP" attempt for the last ~4 months returned HTTP 500. This PR rewrites the clone block against the current models.

**Pair this with**: PR O (admin cascade bypass via auth manager). PR D restricts to "non-admins see what they own; admins see everything." PR O makes "admin" check the Ontos role system instead of just the workspace `admins` group — without it, a customer admin assigned via a non-`admins` workspace group sees an empty Products page.

### Ownership-cascade scoping

`GET /api/data-products` previously returned ALL products to every non-admin caller when no `project_id` query-param was set. `PUT /api/data-products/{id}` (update_product_with_auth) only enforced project membership when the target product had `project_id` set; otherwise no ownership check at all.

This PR applies the same ownership cascade to BOTH list and update paths:

```
admin
 -> project_id ∈ caller's project memberships
 -> owner_team_id ∈ caller's team memberships
 -> draft_owner_id == caller's email
 -> deny (or empty list)
```

Key design choices:
- `draft_owner_id` serves as the "creator / single-user ownership" field for both drafts and published products. Schema already has it indexed and nullable; no migration needed.
- List filter enforced at SQL level via OR'd predicates in the repository — one query, no N+1.
- Update check is in-code in the manager so the cascade can use `projects_manager.is_user_project_member` (which already considers group membership).
- Non-admin callers with no resolvable scope (no email, no teams, no projects) fail closed (empty list / 403). Legacy orphan rows (no project_id, owner_team_id, OR draft_owner_id) are invisible/uneditable for non-admins until an admin promotes them.
- System call sites that legitimately need full visibility (`get_search_index_items`, `get_published_products` marketplace, `save_to_yaml`) pass `is_admin=True` explicitly to bypass the cascade.

### Clone-product schema-drift repair (bundled commit)

The clone routine referenced columns/attributes that no longer exist on the DB models:
- `OutputPortDb`: removed `expectation`, `dataset_name`; correct columns are `description`, `port_type`, `delivery_method_id`, `asset_type`, `asset_identifier`, `status`, `server`, `contains_pii`, `auto_approve`
- `SBOMDb`: only `type` and `url` today (removed `spdx_version`, `spdx_id`, `name`, `creation_info_created`)
- `ManagementPortDb`: requires `content` (was omitted); uses `port_type` / `url` (not `type` / `endpoint`)
- `SupportDb`: requires `channel` (was omitted); uses `tool` / `scope`
- `DataProductTeamMemberDb`: requires `username` (was `email` — no such column)
- The relationship is `output_port.sbom` (singular), not `output_port.sboms` — the `hasattr` check silently no-op'd, so SBOMs were never carried over

Without this fix, `clone_product_for_new_version` raised `AttributeError` on the first output port for any product with ports, or `IntegrityError` on the first management/support/team member if the `AttributeError` didn't fire first.

Bundled here because PR D already heavily modifies `data_products_manager.py` (the same file). The two changes are unrelated semantically (cascade vs clone correctness) but landing them together avoids a second round-trip on the same file in review.

## Safety contracts

- **Stacked on #376** (PR A's branch). Merge order: #376 first.
- **Must ship with PR O** (cascade-bypass for Ontos-role admins): without PR O, customers whose admin role is assigned to a non-`admins` workspace group lose Products page visibility entirely. **Do not merge #393 without also merging the PR O equivalent (filed separately).** The "the customer regression" was exactly this gap.

## Tests

- `test_data_products_ownership_scope.py` — 23 cases at repo + manager layers, admin regression guards, fail-closed orphan handling
- `test_data_products_repository.py` + `test_data_products_manager.py` — list/get_multi tests updated to pass `is_admin=True` since they exercise listing mechanics, not auth
- `test_clone_product_schema_drift.py` — 6 new cases: OutputPort columns, SBOMs via singular relationship, ManagementPort required `content`, Support required `channel`, TeamMember `username`, end-to-end clone of a product with every entity type

23+6 = 29 new tests, all pass.

## Verification

Verified on a production workspaceuction — Products page works for admins (with PR O in the same bundle), Data Producers see scoped lists, clone-for-editing on active DPs now returns 200.

### DIFF
--- a/src/backend/src/controller/data_products_manager.py
+++ b/src/backend/src/controller/data_products_manager.py
@@ -242,15 +242,30 @@ class DataProductsManager(DeliveryMixin, SearchableAsset):
         skip: int = 0,
         limit: int = 100,
         project_id: Optional[str] = None,
-        is_admin: bool = False
+        is_admin: bool = False,
+        caller_email: Optional[str] = None,
+        caller_team_ids: Optional[List[str]] = None,
+        caller_project_ids: Optional[List[str]] = None,
     ) -> List[DataProductApi]:
         """List ODPS v1.0.0 data products.
+
+        For non-admin callers, results are filtered by the ownership cascade
+        described on ``DataProductsRepository.get_multi``: a product is
+        visible if it matches the caller's project membership, team
+        membership, or creator (``draft_owner_id``).
+
+        Back-compat: existing call sites that pass only ``is_admin`` (or
+        nothing) continue to compile. Non-admin calls with no scope inputs
+        intentionally return an empty list — fail-closed.
 
         Args:
             skip: Number of records to skip
             limit: Maximum number of records to return
             project_id: Optional project ID to filter by (ignored if is_admin=True)
-            is_admin: If True, return all products regardless of project_id
+            is_admin: If True, return all products regardless of scope
+            caller_email: Caller email — matched against ``draft_owner_id``
+            caller_team_ids: Caller team memberships — matched against ``owner_team_id``
+            caller_project_ids: Caller's accessible projects — matched against ``project_id``
 
         Returns:
             List of DataProduct API models
@@ -261,7 +276,10 @@ class DataProductsManager(DeliveryMixin, SearchableAsset):
                 skip=skip,
                 limit=limit,
                 project_id=project_id,
-                is_admin=is_admin
+                is_admin=is_admin,
+                caller_email=caller_email,
+                caller_team_ids=caller_team_ids,
+                caller_project_ids=caller_project_ids,
             )
             products_with_tags = []
             for product_db in products_db:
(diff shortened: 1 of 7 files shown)


larsgeorge-db (dismissed):
lgtm

PR #396, opened 2026-05-20
feat(data-products): preserve delivery_method_id on partial update + project_id on create

mvkonchits-db:

## 🇬🇧 Plain English

**Two field-preservation fixes in the data-product repository:**

1. **Before this PR**, every time anyone saved a data product, the output ports' `delivery_method_id` got silently wiped to `None` — even if the user hadn't changed it. This is the field that binds a port to a delivery method (e.g., "this port serves data as a Unity Catalog table"). Customers were re-binding delivery methods after every save without realizing what was happening. This PR preserves the value through saves.

2. **Bundled bug fix**: when creating a data product with a `project_id` set (the project the DP belongs to), the project association was getting silently dropped at the repository layer because the code hard-coded `project_id=None`. POST included the field, the schema accepted it, but the repo threw it away. This PR reads the value from the input instead.

### Output-port delivery-method preservation

The upsert path in `DataProductsRepository.update` iterated all ports and reset every column from the dict — including columns the caller never sent. With PR I's `exclude_unset=True` (a parallel PR), partial updates only carry the keys the caller actually set; pre-PR-I, the route called `model_dump()` with all defaults, so `delivery_method_id=None` was always in the dict and always overwrote the stored value.

This PR adds a `if 'delivery_method_id' in port_dict: ...` guard at the upsert path. Now:
- Caller sends `delivery_method_id: <uuid>` → stored as uuid
- Caller sends `delivery_method_id: null` (explicit clear) → stored as null (Pydantic does NOT strip `None` if the caller included the key)
- Caller omits `delivery_method_id` entirely → existing value preserved (this is the new behavior)

The full effect requires PR I (PUT route uses `exclude_unset=True`); without PR I, the guard is dormant but harmless.

### Bug Q — preserve `project_id` on create

`DataProductsRepository.create()` previously hard-coded `project_id=None` with a stale comment claiming the manager would set it later — which it never did. Any data product POSTed with a `project_id` silently lost the association on create.

Fix: read `project_id` from `DataProductCreate` (the schema declares the field at line 487). No migration needed.

## Safety contracts

- **Pair with PR I**: PR E's guard is dormant until PR I enables partial updates. Both ship together.
- **No dependency on PR D or PR H**.

## Tests

- 5 unit tests for delivery_method_id preservation: set / explicit-null / omit / re-bind
- 1 unit test for project_id preservation on create
- 21 existing repository tests still pass

## Verification

Verified on a production workspaceuction — 12/12 iterations of E2E delivery-method round-trip pass. `a test data product` and `a test data product` (existing customer DPs with delivery method "Table Access" assigned) confirmed to retain their bindings through partial updates.

### DIFF
--- a/src/backend/src/repositories/data_products_repository.py
+++ b/src/backend/src/repositories/data_products_repository.py
@@ -77,7 +77,13 @@ class DataProductRepository(CRUDBase[DataProductDb, DataProductCreate, DataProdu
                 domain=obj_in.domain,
                 tenant=obj_in.tenant,
                 owner_team_id=obj_in.owner_team_id,
-                project_id=None,  # Set via manager if needed
+                # Preserve project_id from the input schema. Historic
+                # comment claimed "Set via manager if needed", but the
+                # manager never populated it from this path, so any DP
+                # POSTed with a project_id silently lost it on create.
+                # (DataProductCreate / DataProduct both declare the
+                # field, so reading via attribute is safe.)
+                project_id=getattr(obj_in, 'project_id', None),
                 max_level_inheritance=obj_in.max_level_inheritance,
                 parent_product_id=getattr(obj_in, 'parent_product_id', None),
                 base_name=getattr(obj_in, 'base_name', None),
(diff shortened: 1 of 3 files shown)


larsgeorge-db (approved):
lgtm

PR #397, opened 2026-05-20
fix(data-products): UUID serializer + missing schema fields on DataProductCreate/Update

mvkonchits-db:

## 🇬🇧 Plain English

**Two model-layer fixes:**

1. **Before this PR**, the API returned `OutputPort.deliveryMethodId` as a Python UUID object via one path (the model-validation path) and as a string via another (the raw-DB path). The frontend expects a string; if it received a UUID, it broke. This PR forces all paths to return a string.

2. **Bundled bug fix**: `DataProductCreate` and `DataProductUpdate` schemas were missing three fields that exist as columns on the database (`draft_owner_id`, `base_name`, `change_summary`). Because Pydantic v2 with default settings silently drops undeclared fields, every POST or PUT carrying these values had them stripped before reaching the repository. This PR declares the fields on both schemas so they round-trip correctly.

### UUID-to-string coercion on `OutputPort.deliveryMethodId`

`OutputPortDb.delivery_method_id` is `PG_UUID(as_uuid=True)` — SQLAlchemy hands us a `uuid.UUID` object. The Pydantic `OutputPort.deliveryMethodId` is typed `Optional[str]`. When constructed from a dict (`OutputPort(**port_dict)`), the type was checked but not coerced — depending on whether the caller passed `str(uuid)` or the raw UUID, the field's value type at runtime differed. JSON serialization handles `UUID` via `default=str` in some paths and not others.

This PR adds `@field_validator('deliveryMethodId', mode='before')` to coerce UUID→str at input, and a defensive `@field_serializer('deliveryMethodId')` to ensure JSON output is always a string. Type contract honored end-to-end.

### Bug R — three missing schema fields

`DataProductCreate` (line 473) and `DataProductUpdate` (line 546) were missing:
- `draft_owner_id` — used for personal-draft ownership; missing it meant clone-for-editing personal drafts had no owner
- `base_name` — stable name across versions; lineage broke for clones
- `change_summary` — version-changelog string; never persisted

`model_config` doesn't set `extra="allow"`, so Pydantic v2 silently drops undeclared keys. The repository's defensive `getattr(obj_in, 'draft_owner_id', None)` always returned `None`. The repo defaults were used instead of the caller's intent.

Bug R declares all three on both schemas. Columns already exist on `DataProductDb` — no migration needed.

## Safety contracts

- **No PR dependency**.
- **Pairs implicitly with PR I**: PR I's `exclude_unset=True` makes the schema-declared fields round-trip correctly. Bug R ensures the fields are declared.

## Tests

- 8 unit tests for missing-fields round-trip on Create and Update (each of the 3 fields, plus exclude_unset semantics for clearing draft_owner_id distinguishable from omitting it)
- Existing UUID-coercion tests pass

## Verification

Verified on a production workspaceuction — output ports with delivery method assignments return correctly typed strings to the UI; data products created via the wizard correctly carry their `draft_owner_id` and `change_summary` into the DB.

### DIFF
--- a/src/backend/src/models/data_products.py
+++ b/src/backend/src/models/data_products.py
@@ -497,6 +497,15 @@ class DataProductCreate(BaseModel):
 
     # Versioning
     parent_product_id: Optional[str] = Field(None, alias="parentProductId", description="Parent version ID for version lineage")
+    # The following three columns exist on DataProductDb but were missing
+    # from the create schema, so Pydantic v2 silently stripped them from
+    # incoming POSTs (model_config does not set extra="allow"). The
+    # repository's getattr() fallback then returned None on every
+    # create. Declaring them here makes the create round-trip honor
+    # the input. No DB migration needed — columns already exist.
+    draft_owner_id: Optional[str] = Field(None, description="Creator / single-user owner email for personal drafts and non-team-owned products")
+    base_name: Optional[str] = Field(None, description="Stable base name shared across versions of the same product family")
+    change_summary: Optional[str] = Field(None, description="Free-text summary of changes in this version")
 
     # Metadata inheritance
     max_level_inheritance: int = Field(99, ge=0, le=999, description="Maximum metadata level to inherit from contracts")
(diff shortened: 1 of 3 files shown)


larsgeorge-db (approved):
lgtm

PR #398, opened 2026-05-20
fix(data-products): exclude_unset on PUT preserves unmodified Optional fields

mvkonchits-db:

## 🇬🇧 Plain English

**Before this PR**, when you saved a data product (PUT to `/api/data-products/{id}`), every Optional field you DIDN'T explicitly include in your request got silently wiped to `None`. So if you only wanted to change the name and didn't send `description`, `delivery_method_id`, `contract_id`, etc. in the body, they all got cleared. UI partial-save flows became impossible without re-sending the entire DP every time.

**Root cause**: the route called `product_update.model_dump()` (without `exclude_unset=True`). Pydantic dumps EVERY field with its default value — so the dict reaching the manager had `description=None`, `delivery_method_id=None`, etc. for every absent field. The repo's `if 'field' in update_data: db_obj.field = ...` pattern then dutifully overwrote every column to `None`.

**Fix**: `product_update.model_dump(exclude_unset=True)` — the dict only contains fields the caller actually sent. Repository iterates only what's in the dict, leaves the rest alone.

## Technical detail

The change is one argument on one line in `data_product_routes.py` (around line 1963), but the empirical diagnosis took a focused Pydantic-rebuilding experiment to confirm: `model_dump()` is well-documented but the route's intent (partial update) wasn't matched by its implementation. The bug had been there since the `DataProductUpdate` model was added.

The repository layer (`DataProductsRepository.update`) was already correct in its iteration shape — it checks `if 'field' in update_data` before touching each column. So once the dict is shaped correctly (only contains caller-supplied keys), the existing repo code does the right thing.

## Safety contracts

- **MUST pair with PR E and PR H** — without PR E's preservation guard for `delivery_method_id` and PR H's declared `draft_owner_id/base_name/change_summary`, partial updates would still wipe those fields (because they wouldn't be in the dict, but they also wouldn't have preserve logic). With this triple, partial updates work correctly for ALL Optional fields, no exceptions.
- **No DB migration**.

## Tests

- 1 integration test for partial PUT preservation: send `{name: "new name"}` only, expect description / port / delivery_method_id all retained
- Existing PUT tests still pass

## Verification

Verified on a production workspaceuction — 3/3 iterations of partial-PUT round-trip pass.

### DIFF
--- a/src/backend/src/routes/data_product_routes.py
+++ b/src/backend/src/routes/data_product_routes.py
@@ -1891,7 +1891,13 @@ async def update_data_product(
 
         # Delegate to manager (includes auth check)
         user_groups = current_user.groups or []
-        product_dict = product_update.model_dump()
+        # PR I: exclude_unset preserves partial-update semantics. Without it,
+        # Pydantic's defaults flood the dict with None for every Optional field,
+        # the manager re-instantiates DataProductUpdate(**full_dump) marking all
+        # fields as "set", and the repository's
+        # `if 'field' in update_data: db_obj.field = ...` pattern clears every
+        # unmodified Optional column (delivery_method_id, contract_id, etc).
+        product_dict = product_update.model_dump(exclude_unset=True)
 
         updated_product_response = manager.update_product_with_auth(
             product_id=product_id,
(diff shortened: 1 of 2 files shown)


larsgeorge-db (approved):
lgtm

Issue #335, opened 2026-05-06
[PRD]: Entra ID / Microsoft Graph integration & unified Principal Picker


larsgeorge-db:
## Summary

Add an **optional** Entra ID (Azure AD) integration via a Unity Catalog HTTP Connection, plus a unified `PrincipalPicker` component that replaces ~14 inconsistent principal-entry UIs across the app with a Gmail-style badge input.

- **Configured mode**: type-ahead + popup search dialog backed by Microsoft Graph (Users + Groups). Each result row shows `displayName` on top and email/UPN (users) or GUID (groups) underneath, so homonyms are always disambiguated.
- **Unconfigured mode**: typed values become removable badges; clicking a badge reverts it to editable text.
- **Auth**: UC HTTP Connection only — UC handles OAuth2 client credentials and token refresh via `ws.serving_endpoints.http_request(conn=name, ...)`. **No app-side token cache, no client secret stored in the app database.**
- **Storage**: keep existing `string` / `List[str]` fields untouched. No Alembic migration for v1.

Full PRD: [`docs/prds/prd-entra-id-graph-integration.md`](../blob/feat/data-catalog-333/docs/prds/prd-entra-id-graph-integration.md)

## Why

Today, ~14 different forms each implement principal entry differently (plain `Input`, comma-separated string, hard-coded `Select`, even unwired wizard placeholders). There is no validation, no display-name resolution, and no defense against typos. Operators with an Entra tenant cannot pick from a directory; consumers are guessing whether a group is `data-eng` or `data_engineering`.

The existing UC HTTP Connection plumbing (used by the workflow webhook step) already gives us a clean, secret-free path to Microsoft Graph. The pieces are in place; what is missing is one shared component, a small Graph search backend, and migration of every call site.

## Decisions locked in (from PRD Q&A)

- Replacement scope: **all ~14 call sites** in v1.
- Principal types: **Users + Groups** (defer service principals).
- Storage: **keep strings** (`List[str]` of emails / group names).
- Search UI: **both** type-ahead AND popup dialog in the same component.
- Pre-existing data: **no re-validation**; renders as plain badges.
- Manual-mode badges: **click to edit** (true Gmail-like).
- Disambiguation: **two-line rows** (name + email/GUID) plus tooltip on selected badges.

### Backend
- [ ] Extract shared `list_http_connections(ws)` helper from `workflows_routes.py` and expose `GET /api/settings/uc-http-connections`
[...]

PR #406, opened 2026-05-21
feat(directory): Phase 1 backend — Directory layer + Entra provider

larsgeorge-db:
First of two PRs implementing Phase 1 of the directory-lookup-and-principal-picker plan (#375). This PR is the backend foundation; a follow-up PR will add the Settings tab + `PrincipalPicker` + Assign Owner migration.

Source PRD: [docs/prds/prd-entra-id-graph-integration.md](https://github.com/databrickslabs/ontos/blob/main/docs/prds/prd-entra-id-graph-integration.md) (#335)
Plan: [plans/directory-lookup-and-principal-picker.md](https://github.com/databrickslabs/ontos/blob/plan/directory-lookup-and-principal-picker/plans/directory-lookup-and-principal-picker.md) (#375)

## Summary
- New `Directory` layer: provider-agnostic abstraction (`DirectoryProvider` ABC) with a single concrete provider for Microsoft Entra ID. All Graph traffic goes through `ws.serving_endpoints.http_request(connection_name=...)` so UC owns OAuth2 client-credentials acquisition and token caching — the app stores no secret.
- `DirectoryManager` dispatches to the configured provider and caches results in a 5-min in-memory TTL keyed on `(provider_type, connection_name, kind, query|limit)`, invalidating on settings change or explicit `invalidate_cache()`.
- Settings live in the existing `app_settings` key/value table (`DIRECTORY_PROVIDER_TYPE`, `DIRECTORY_UC_HTTP_CONNECTION_NAME`) — **no Alembic migration**.
- Five routes under `/api/directory/*`: `status`, `search`, `test`, `settings`, `uc-http-connections`. Permission-gated against the existing `settings` feature.
- Shared `list_http_connections()` helper extracted from `workflows_routes.py`; the existing `/api/workflows/http-connections` route now delegates.
- Test conftest gains non-secret `setdefault`s for required Settings fields so the suite is runnable locally without `.env`.

## What's intentionally **not** here
- Frontend (`PrincipalPicker`, Settings → Directory tab, Assign Owner migration) — next PR.
- Production second provider — the abstraction is exercised by a unit-test stub. Okta / Ping land in a future phase per plan's out-of-scope list.
- Storage-shape changes — every existing principal-bearing field keeps its string / `List[str]` shape.

## Architectural notes
- `Principal.id` is the **persisted identifier** (UPN/email for users, displayName for groups — *not* GUIDs). `sub_label` carries the secondary identifier (email/UPN or GUID) for disambiguation.
- Unknown provider types in settings are treated as not-configured. UI uses `DirectoryStatus.configured` to decide picker mode.
- All search responses are de-duplicated by `(type, id)` to survive partial cache hits and the cross-type merge honours the caller's overall `limit`.

## Test plan
- [x] Backend unit suite: `cd src && hatch -e dev run pytest backend/src/tests/unit/` → **764 passed, 0 failed**
- [x] New tests:
  - `test_entra_id_provider.py` (16): OData escaping, `$select` projections, Principal mapping (UPN → id for users, displayName → id for groups), eventual-consistency header, response body shapes (bytes / str / stream / None), Graph error body surfaced as `DirectoryError`, transport-error surfaces, non-JSON body, empty connection-name guard.
  - `test_directory_manager.py` (14): unknown-provider-type stays not-configured, dispatches to registered provider, cache hits on case + whitespace normalisation, cache invalidates when settings change or `invalidate_cache()` is called, `types=["user"]` narrows downstream calls, **stub-provider registration proves the abstraction is enough to add a provider with no manager/route changes**.
- [x] Lint clean for new modules (`ruff check --isolated`).
- [ ] Manual smoke test of `/api/directory/status` against a workspace (left for reviewer or follow-up PR alongside the UI).

### DIFF
--- a/src/backend/src/app.py
+++ b/src/backend/src/app.py
@@ -74,6 +74,7 @@ from src.routes import (
     readiness_routes,
     suggestion_routes,
     certification_levels_routes,
+    directory_routes,
 )
 
 from src.common.database import init_db, get_session_factory, SQLAlchemySession
@@ -383,6 +384,7 @@ mcp_tokens_routes.register_routes(app)
 self_service_routes.register_routes(app)
 workflows_routes.register_routes(app)
 settings_routes.register_routes(app)
+directory_routes.register_routes(app)
 connection_routes.register_routes(app)
 schema_import_routes.register_routes(app)
 
--- a/src/backend/src/common/dependencies.py
+++ b/src/backend/src/common/dependencies.py
@@ -25,6 +25,7 @@ from src.controller.business_roles_manager import BusinessRolesManager
 from src.controller.business_owners_manager import BusinessOwnersManager
 from src.controller.delivery_methods_manager import DeliveryMethodsManager
 from src.controller.ontology_generator_manager import OntologyGeneratorManager
+from src.controller.directory_manager import DirectoryManager
 
 # Import base dependencies
 from src.common.database import get_session_factory # Import the factory function
@@ -55,6 +56,7 @@ from src.common.manager_dependencies import (
     get_business_owners_manager,
     get_delivery_methods_manager,
     get_ontology_generator_manager,
+    get_directory_manager,
 )
 # Import workspace client getter separately as it might be structured differently
 from src.common.workspace_client import get_workspace_client_dependency  # Fixed to use proper wrapper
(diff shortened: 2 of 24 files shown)


mvkonchits-db (dismissed):
LGTM — approving. Tested locally as part of the full stack and the abstraction holds.

Verified separately: UC owns OAuth2 entirely, app stores no tokens/secrets (only the non-secret `DIRECTORY_PROVIDER_TYPE` and `DIRECTORY_UC_HTTP_CONNECTION_NAME` keys in `app_settings`). OData single-quote escaping in `_escape_odata` is correct; path-segment lookups go through `_url_quote` which is the right tool for that side. Permission gating on the 5 new routes is consistent (`settings:READ_ONLY` for status/search, `settings:READ_WRITE` for test/settings/uc-http-connections) — end users get `search` + `status`, which is what the picker needs.

**Two small nits.**

`_user_to_principal` (`entra_id_provider.py:573-587`) has a `sub_label` dedup branch that reads as accidentally-tautological: `sub_label = u.get("userPrincipalName") if sub_label != u.get("userPrincipalName") else None`. Works as intended, but the next reader will pause — worth either a clarifying comment or a rewrite to make the intent obvious.

The in-memory 5-min TTL cache lives on the manager instance (`app.state`). With multiple Uvicorn workers each worker maintains its own cache, so settings changes are eventually-consistent across workers within the TTL. Fine for the current shape, just worth a docstring note so the next person doesn't assume cross-worker invalidation.

PR #407, opened 2026-05-21
feat(directory): Phase 1 frontend — Settings tab + PrincipalPicker

larsgeorge-db:
Second of two PRs implementing Phase 1 of the directory-lookup-and-principal-picker plan (#375). **Stacked on #406** — base branch is `feat/directory-backend`, so the diff here is frontend-only. Will rebase onto `main` once #406 merges.

Source PRD: #335
Plan: #375
Backend PR: #406

## Summary
- **Settings → Integrations → Directory** tab: provider Select (Entra ID enabled, Okta/Ping rendered disabled so the abstraction is visible to operators), UC HTTP Connection dropdown fed from `/api/directory/uc-http-connections`, Save / Test / Clear buttons with success/destructive toast feedback, and a provider-specific help block for Entra with the four UC-connection fields plus required Graph application scopes.
- **`PrincipalPicker`** — single component, both modes (configured / unconfigured) and both UI variants (inline 2-char-debounced type-ahead + popup "Browse directory" dialog with type-filter chips) selected at runtime from the directory store. Two-line result rows (display name + sub_label), tooltip on selected badges exposing the underlying email/UPN/GUID, X-to-remove. Props are discriminated on `multiple` so single-pick callers get `string | null` typing and multi-pick callers get `string[]` typing without casts.
- **`directory-store` (Zustand)** — singleton cache for `/api/directory/status` so every picker on a page shares one network call, plus a session-sticky `degraded` flag flipped on the first search failure so subsequent pickers drop straight into manual mode without re-probing.
- **Assign Owner dialog** migrated to `PrincipalPicker(accepts=['user'])`. API payload (`user_email` + `user_name`) is unchanged — the change is purely UI.
- Sidebar entry under Settings → Integrations and `/settings/directory` route wired up.

## Architectural notes
- The picker emits and consumes only the persisted identifier string (UPN/email for users, displayName for groups). Existing storage shapes (`str` / `List[str]`) are preserved; no DB migration in v1.
- Pre-existing values render as plain badges with no error decoration regardless of provider availability. The picker does *not* re-resolve persisted ids against the directory.
- Graceful degradation: if `directory.status` says configured but a search call fails, the picker logs once (`console.warn`), flips into manual-entry mode for the rest of the session, and `directory-store.degraded` becomes sticky so newly-mounted pickers behave the same way.
- Future Phase-2+ migrations (Roles, Entitlements, Comments, etc.) reuse this same component with different `multiple` / `accepts` combinations — no per-call-site bespoke control.

## Test plan
- [x] Type-check clean: `yarn type-check` (Done in 17.93s)
- [x] Lint clean on new files (`npx eslint ...`): **0 errors**, only the repo-wide `err: any` catch-block warning pattern remains (matches existing files).
- [x] Frontend suite: `npx vitest run` → **550 passed, 6 skipped, 0 failed**.
- [x] New tests (19):
  - `principal-picker.test.tsx` (13): pre-existing values render as plain badges without re-resolving; X-remove emits remaining ids; disabled hides remove buttons; Enter/Tab/comma/blur commit in unconfigured mode; empty / duplicate inputs are no-ops; configured mode calls `/api/directory/search` with `q=` and `types=` honouring `accepts`; queries < 2 chars do not fire a request; failing search flips into manual-entry mode for the rest of the session.
  - `directory-store.test.ts` (6): `fetchStatus` populates state, dedupes concurrent calls into one in-flight request, no-ops once loaded, falls back to not-configured on network failure; `refresh` re-fetches and clears `degraded`; `markDegraded` is sticky.
- Radix Popover dropdown internals are deferred to Playwright E2E — the existing `tag-selector` tests in this repo follow the same split for the same reason (Radix Popover hangs in jsdom).
- [ ] Manual smoke test of the Settings tab + Assign Owner against a real workspace with an Entra UC connection (left for reviewer / follow-up).

## What's intentionally **not** here
- Migration of other principal-bearing fields (Roles, Entitlements, Comments, Teams, Reviews, MDM, etc.) — Phases 2–4 in the plan.
- A production second provider — abstraction is exercised by backend unit-test stub.

### DIFF
--- a/src/backend/src/app.py
+++ b/src/backend/src/app.py
@@ -74,6 +74,7 @@ from src.routes import (
     readiness_routes,
     suggestion_routes,
     certification_levels_routes,
+    directory_routes,
 )
 
 from src.common.database import init_db, get_session_factory, SQLAlchemySession
@@ -383,6 +384,7 @@ mcp_tokens_routes.register_routes(app)
 self_service_routes.register_routes(app)
 workflows_routes.register_routes(app)
 settings_routes.register_routes(app)
+directory_routes.register_routes(app)
 connection_routes.register_routes(app)
 schema_import_routes.register_routes(app)
 
--- a/src/backend/src/common/dependencies.py
+++ b/src/backend/src/common/dependencies.py
@@ -25,6 +25,7 @@ from src.controller.business_roles_manager import BusinessRolesManager
 from src.controller.business_owners_manager import BusinessOwnersManager
 from src.controller.delivery_methods_manager import DeliveryMethodsManager
 from src.controller.ontology_generator_manager import OntologyGeneratorManager
+from src.controller.directory_manager import DirectoryManager
 
 # Import base dependencies
 from src.common.database import get_session_factory # Import the factory function
@@ -55,6 +56,7 @@ from src.common.manager_dependencies import (
     get_business_owners_manager,
     get_delivery_methods_manager,
     get_ontology_generator_manager,
+    get_directory_manager,
 )
 # Import workspace client getter separately as it might be structured differently
 from src.common.workspace_client import get_workspace_client_dependency  # Fixed to use proper wrapper
@@ -132,6 +134,7 @@ BusinessRolesManagerDep = Annotated[BusinessRolesManager, Depends(get_business_r
 BusinessOwnersManagerDep = Annotated[BusinessOwnersManager, Depends(get_business_owners_manager)]
 DeliveryMethodsManagerDep = Annotated[DeliveryMethodsManager, Depends(get_delivery_methods_manager)]
 OntologyGeneratorManagerDep = Annotated[OntologyGeneratorManager, Depends(get_ontology_generator_manager)]
+DirectoryManagerDep = Annotated[DirectoryManager, Depends(get_directory_manager)]
 
 # Permission Checker Dependency
 PermissionCheckerDep = AuthorizationManagerDep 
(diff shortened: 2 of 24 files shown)


mvkonchits-db (approved):
LGTM — approving. Settings tab + PrincipalPicker are clean; tested locally end-to-end as part of the full stack.

Verified the `directory-store` is a real singleton (one `/api/directory/status` fetch per page, dedupes concurrent calls, `degraded` flag is session-sticky). Assign Owner dialog payload (`user_email` + `user_name`) is unchanged.

**Two nits.**

The `multiple` prop is meant to discriminate the value type (`string | null` vs `string[]`), but the picker internally does `props as PrincipalPickerProps & { multiple?: boolean }` and `emit()` re-casts at the same spots. Works correctly at runtime — the discriminated union just isn't doing the type-narrowing job it was set up for.

The `useMemo`/`useEffect` deps use `accepts.join(',')` with `eslint-disable react-hooks/exhaustive-deps`. Works as long as callers pass a stable `accepts` reference, but it's a footgun for the next site that inlines `accepts={['user']}` in a render path.

**Follow-up (whole stack, not this PR).**
Non-admin personas opening any surface with a picker will see a 403 on `/api/directory/status` because the routes are gated on `settings:READ_ONLY`. Functionally safe (picker degrades to manual entry), but noisy in network panels. Worth tracking — either widen `status`+`search` to a non-`settings` feature read scope or add a tiny `directory:READ_ONLY` feature defaulted broadly.

PR #411, opened 2026-05-21
docs(contributing): document private npm mirror lockfile workflow

larsgeorge-db:
## Summary

Adds an optional **Working Behind a Private npm Mirror** subsection to `CONTRIBUTING.md` under frontend setup. Targets contributors whose corporate networks force all npm/yarn traffic through an internal mirror (Nexus, Artifactory, Verdaccio, HTTPS proxy, etc.), which would otherwise rewrite every `resolved` URL in `src/frontend/yarn.lock` from `registry.yarnpkg.com` to the mirror host and produce huge unwanted diffs on every `yarn install`.

The docs walk through a per-clone, **local-only** git clean/smudge filter setup:

- `filter.npmmirror.clean` rewrites mirror URLs back to `registry.yarnpkg.com` on `git add`
- `filter.npmmirror.smudge` rewrites the other way on checkout
- Attached via `.git/info/attributes` (never committed), so external contributors are completely unaffected
- Renormalize the working tree once, then `yarn install --frozen-lockfile`, `yarn add`, and `yarn upgrade` all work normally and commits stay OSS-clean

Generic phrasing — no specific mirror host is named, uses `<your-mirror-host>` as placeholder.

## Test plan

- [x] `yarn install --frozen-lockfile` succeeds with mirror URLs on disk
- [x] `git diff src/frontend/yarn.lock` is empty after install
- [x] `git show :src/frontend/yarn.lock` contains only `registry.yarnpkg.com` URLs
- [ ] Reviewer skims the new subsection in `CONTRIBUTING.md` for clarity

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -65,6 +65,59 @@ yarn install
 # Backend dependencies are managed by Hatch and installed automatically
 ```
 
+#### Working Behind a Private npm Mirror (optional)
+
+Some corporate environments block the public npm/yarn registry and require all
+package traffic to flow through an internal mirror (e.g., Nexus, Artifactory,
+Verdaccio, or an HTTPS pass-through proxy). If that applies to you, running
+`yarn install` will rewrite every `resolved` URL in `src/frontend/yarn.lock`
+from `https://registry.yarnpkg.com/...` to your mirror's URL, producing a huge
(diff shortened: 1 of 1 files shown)


will-yuponce-db (approved):

will-yuponce-db:
Looks like there's an unrealated failing tests. Fixing it then I'll rebase

PR #413, opened 2026-05-21
feat(directory): Phase 3 — Reviews / MDM / Tags / Teams / DP+DC team-member

larsgeorge-db:
Phase 3 of the directory-lookup-and-principal-picker plan (#375). **Stacked on #412** — base is `feat/directory-phase2` so the diff here is Phase-3 only. Rebases onto `main` once the rest of the stack lands.

Plan: #375
Backend: #406
Phase 1 frontend: #407
Phase 2: #412

## Summary
Five mechanical migrations onto `PrincipalPicker`. No new backend routes; every existing endpoint accepts the same payload shape it already did.

| Surface | Field | Picker config | Backend payload |
|---|---|---|---|
| Data Asset Reviews — Create | `reviewer_email` | single, user | unchanged |
| MDM — Create Review | `reviewer_email` | single, user (RHF `Controller`) | unchanged |
| Tag-namespace ACL | `permissionForm.group_id` | single, group | unchanged |
| Teams form | `member_identifier` | single, **dynamic** — `accepts` flips between `['user']` and `['group']` on the row's `member_type` toggle | unchanged |
| Data-product team-member | `username` | single, user | unchanged (`username`) |
| Data-contract team-member | email/username | single, user | unchanged (still populates **both** `username` and `email` for ODCS backward compat) |

## Helpers added
- `lib/team-members.ts`:
  - `principalAcceptsForMemberType(memberType)` — Teams type-switching.
  - `buildContractTeamMember({emailOrUsername, role, name})` — DC dual-population helper (trims, omits `name` when blank, populates both `username` and `email` with the same picked id).

## Test plan
- [x] Type-check clean (`yarn type-check`).
- [x] Full frontend suite: **705 passed, 6 skipped, 0 failed**.
- [x] New tests (6, `lib/team-members.test.ts`):
  - `principalAcceptsForMemberType`: `'user'` → `['user']`, `'group'` → `['group']`, defaults to `['user']` for `null` / `undefined` / `''` / unrecognised values.
  - `buildContractTeamMember`: populates both `username` and `email` with the picked id; trims surrounding whitespace on every field; omits `name` entirely when blank or undefined.
- Per-page form-dialog mounts continue to be deferred to Playwright E2E (Radix dialogs hang in jsdom — same convention as the existing skipped `role-form-dialog.test`, `team-form-dialog.test`, `project-form-dialog.test`, `tag-selector.test`).
[...]

PR #412, opened 2026-05-21
feat(directory): Phase 2 — Roles + Entitlements + Comments

larsgeorge-db:
Phase 2 of the directory-lookup-and-principal-picker plan (#375). **Stacked on #407** — base is `feat/directory-frontend` so the diff here is Phase-2 only. Will rebase onto `main` once #406 and #407 land.

Plan: #375
Backend: #406
Phase 1 frontend: #407

## Summary
Three multi-principal RBAC surfaces migrated to `PrincipalPicker`. No backend changes — every existing endpoint accepts the same payload shape it already did.

- **Roles form** (`role-form-dialog.tsx`): `assigned_groups` comma-separated Input → `PrincipalPicker(multi, groups)`. Legacy comma-string records still rehydrate via the new `normaliseAssignedGroups` helper.
- **Entitlements page** (`entitlements.tsx`): persona-group "add" `Select` with a hard-coded 14-name demo list → `PrincipalPicker(multi, groups)`. The picker renders its own selected-badges so the separate Badge row was dropped. `PUT /api/entitlements/personas/{id}/groups` payload unchanged.
- **Comment composer** (`comment-sidebar.tsx`): audience gains `PrincipalPicker(multi, users+groups)` **alongside** (not replacing) the existing Teams + App Roles checkbox lists. Picked users emit as plain emails, picked groups as plain group names; `team:*` / `role:*` tokens flow through unchanged. The round-trip is extracted into `buildAudienceTokens` / `parseAudienceTokens`.

## Helpers added
- `lib/audience.ts` — `parseAudienceTokens(audience)` / `buildAudienceTokens(parsed)` for the mixed-token round-trip.
- `lib/assigned-groups.ts` — `normaliseAssignedGroups(raw)` tolerates both `string[]` and legacy comma-string shapes.

## Test plan
- [x] Type-check clean (`yarn type-check`).
- [x] Lint clean on touched files (0 errors; only pre-existing repo-wide warnings remain).
- [x] Full frontend suite: **562 passed, 6 skipped, 0 failed**.
- [x] New tests (12):
  - `lib/audience.test.ts` (6): split team:/role:/principal mix, defensive guards for null/empty/non-string, prefix-only tokens preserve their empty id, round-trip identity.
  - `lib/assigned-groups.test.ts` (6): array passthrough, empty-string drop, comma-split + trim, trailing/repeated commas, nullish / non-string-non-array → [], empty string → [].
[...]

PR #416, opened 2026-05-21
feat(directory): add Lakebase + File providers

larsgeorge-db:
Two additional concrete providers + provider-abstraction cleanup. **Stacked on #413** — base is \`feat/directory-phase3\` so the diff here is providers-only. Will rebase onto \`main\` once the rest of the stack lands.

Plan: #375 · Backend (Phase 1): #406 · Phase 1 frontend: #407 · Phase 2: #412 · Phase 3: #413

## Summary
- **\`LakebaseProvider\`** — reads principals from a Postgres / Lakebase table via the app's existing SQLAlchemy engine. Identifier validation rejects SQL-injection attempts at the table-name level; queries use parameterised \`LIKE\`/\`ESCAPE\` for portable case-insensitive prefix search; user wildcards (\`%\`/\`_\`) are escaped so a raw \`%\` can't dump the directory.
- **\`FileProvider\`** — CSV-backed, primarily for tests and offline dev. Required columns: \`type\`, \`id\`, \`display_name\` (\`sub_label\` optional). Re-reads on \`mtime\` change via a class-level cache, no restart needed.
- **Provider abstraction generalisation**: factories now take \`(DirectoryProviderContext, DirectoryProviderConfig)\` instead of \`(ws_client, connection_name)\`. The context carries transport handles (ws_client, db_engine); the config is one bag containing every directory setting and each provider reads only the fields relevant to its type. Adding a future provider is one entry in \`_PROVIDER_REGISTRY\` + an optional \`_REQUIRED_KEYS\` row.
- **Settings keys**: \`DIRECTORY_LAKEBASE_TABLE\`, \`DIRECTORY_FILE_PATH\` added alongside the existing \`DIRECTORY_UC_HTTP_CONNECTION_NAME\`. All three live in \`app_settings\` — still no Alembic migration.
- **\`DirectoryStatus\`** carries all three config fields so the Settings tab hydrates inputs in one round trip.
- Each provider's required-key set is declared centrally; \`configured\` is False whenever any required key for the active provider is missing.

## Settings tab UI
Provider Select enables all three options; the panel below switches between \`EntraPanel\` / \`LakebasePanel\` / \`FilePanel\` on selection. Each renders:
- Entra: UC HTTP Connection dropdown + token/scope/grant-type help block (unchanged from #407).
- Lakebase: FQN \`Input\` + required-schema SQL snippet in the help block.
- File: absolute path \`Input\` + CSV format example.

Form state is preserved per-field when toggling providers so users don't lose typed values.

## Test plan
- [x] Backend directory tests: **67 passed** (16 Entra + 17 manager + 18 Lakebase + 16 File).
- [x] Targeted directory + workflow regression: **158 passed**.
- [x] Frontend type-check clean (\`yarn type-check\`).
- [x] Frontend lint clean on touched files.
- [x] Frontend suite unaffected: **705 passed, 6 skipped**.
- [x] New backend tests:
  - \`test_lakebase_provider.py\` (18): FQN validation rejects illegal identifiers / SQL-injection probes; prefix search hits \`display_name\` and \`id\`; result shape; type filter; \`top\` cap; \`%\`/\`_\` wildcard escape; \`get_user\` happy/missing/empty paths; \`test()\` probe succeeds and fails on missing table; ctx/config validation.
  - \`test_file_provider.py\` (15): CSV parsing happy path, blank rows skipped, blank id rejected, unknown type rejected, missing required columns rejected; case-insensitive prefix; id-column match; \`top\` cap; mtime-based cache re-read; \`test()\` probe.
  - \`test_directory_manager.py\` (+3 new): \`DirectoryStatus\` exposes all per-provider fields; \`configured=False\` when required setting absent; \`manager.test\` raises a clear "missing required" message when a recognised provider lacks its required setting.
  - \`test_entra_id_provider.py\` (16): unchanged behaviour, migrated to the new \`(ctx, config)\` factory signature.
- [ ] Manual smoke test of all three providers in a workspace (left for reviewer / follow-up).

## Notes for reviewers
- The change to the provider factory signature is intentionally invasive — the old \`(ws_client, connection_name)\` couldn't cleanly carry per-provider config. The new shape is the load-bearing piece of "the abstraction was correct".
- \`LakebaseProvider\` uses the app's *own* SQLAlchemy engine via \`db.get_bind()\`. The principal table is expected to live on the same Postgres / Lakebase database the app already talks to. If a customer wants the principal directory on a separate database, that's a future extension (new ctx field).
- \`FileProvider\` is intentionally tiny: no remote fetch, no schema beyond four columns. The intended use is tests, demos, and air-gapped dev environments.

### DIFF
--- a/src/backend/src/controller/directory_manager.py
+++ b/src/backend/src/controller/directory_manager.py
@@ -2,17 +2,24 @@
 
 Reads provider configuration from ``app_settings``, dispatches to the
 right concrete ``DirectoryProvider``, and caches search results in
-memory for 5 minutes per (provider_type, query_shape) key. The cache
-is per-instance and the manager is held as a singleton on
+memory for 5 minutes per (provider_type, config_signature, query) key.
+The cache is per-instance and the manager is held as a singleton on
 ``app.state``.
 
 The manager itself is provider-agnostic: adding a new provider is a
-matter of registering another class in ``_PROVIDER_REGISTRY``.
+matter of:
+
+1. Implementing ``DirectoryProvider`` in ``src.controller.directory_providers``.
+2. Registering it in ``_PROVIDER_REGISTRY`` below.
+3. (If the provider needs new settings) adding the keys to
+   ``DirectoryProviderConfig`` and to the read/write paths here.
+
+No changes to routes or models otherwise.
 """
 
 import time
 from threading import Lock
-from typing import Any, Callable, Dict, List, Optional, Tuple
+from typing import Callable, Dict, List, Optional, Tuple
 
 from sqlalchemy.orm import Session
 
@@ -20,13 +27,19 @@ from src.common.logging import get_logger
 from src.controller.directory_providers import (
     DirectoryError,
     DirectoryProvider,
+    DirectoryProviderConfig,
+    DirectoryProviderContext,
     EntraIdProvider,
+    FileProvider,
+    LakebaseProvider,
 )
 from src.models.directory import (
     DirectoryProviderType,
     DirectoryStatus,
     Principal,
     SETTING_KEY_CONNECTION_NAME,
+    SETTING_KEY_FILE_PATH,
+    SETTING_KEY_LAKEBASE_TABLE,
     SETTING_KEY_PROVIDER_TYPE,
 )
 from src.repositories.app_settings_repository import app_settings_repo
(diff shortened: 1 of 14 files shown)


mvkonchits-db (approved):
LGTM — approving. Tested locally as part of the full stack.

`LakebaseProvider` is solid: FQN regex `^[A-Za-z_][A-Za-z0-9_]*$` per dotted part (max 3 parts) — rejects backticks, semicolons, quotes, dots-inside-segment, unicode lookalikes. Column values bound via `:kind`/`:p`/`:n`/`:id`, LIKE wildcards (`%`, `_`, `\`) escaped with an explicit `ESCAPE '\\'` clause. Real protection, not theatre.

Provider factory signature change `(ws_client, connection_name) → (ctx, config)` ripples cleanly — walked all call sites (`EntraIdProvider`, manager `_build_provider`, test register-provider helper, route-side `_build_context`). No stragglers.

Settings stay in `app_settings` k/v — no Alembic migration, confirmed.

**One concern + one nit + one question.**

`FileProvider` reads `DIRECTORY_FILE_PATH` straight into `open()` with no path-traversal guard. Admin-gated (settings:READ_WRITE) so contained, and CSV header validation rejects non-conformant files, but a malicious admin can still probe file existence and readability via error messages. Suggest at minimum a docstring note that the path is operator-trusted; ideally a `realpath` + denylist of `/etc`, `/proc`, `/sys` or a soft prefix allowlist. Not blocking, but if any customer turns this on in prod the insider-risk surface grows.

`DirectorySettingsUpdate` docstring says "any field may be `None` to clear" but the route treats `None` as skip — only empty-string clears (`body.X or None`). Defensible, but the docstring contradicts the behaviour. Pick one.

**Open question for you, Lars:** is `FileProvider` intended for prod customers (air-gapped Lakebase-less deploys?) or strictly for tests/demos/offline dev? If it's prod-facing the path-traversal hardening becomes more important; if it's test-only, gating its registration behind an env flag might be worth it so a customer admin can't accidentally point it at the wrong place.

`_is_configured` rebuilds the `key_to_field` dict on every call — trivial perf, but it's a static map; lift to module scope.

PR #417, opened 2026-05-21
feat(directory): Phase 4 — Workflow Designer custom principals + DC wizard wiring

larsgeorge-db:
Phase 4 of the directory plan (#375). **Stacked on #416** — base is \`feat/directory-providers\` so the diff is Phase 4 only.

Plan: #375 · Backend (Phase 1): #406 · Phase 1 frontend: #407 · Phase 2: #412 · Phase 3: #413 · Providers: #416

### Workflow Designer — Custom principals toggle
- Notification step and Approval step config panels each gain a "Custom principals" toggle.
  - **Off** = today's role-only \`Select\` (no regression).
  - **On** = adds a \`PrincipalPicker(multi, users+groups)\` alongside; picks join the existing \`recipients\` / \`approvers\` field as comma-separated literals via the new \`joinRoleAndPrincipals\` helper.
- \`splitRoleAndPrincipals\` is the inverse so the designer hydrates the form losslessly from a persisted config. Role-shaped tokens (\`requester\`, \`owner\`, legacy aliases like \`domain_owners\`, UUID-shaped IDs, \`business:<uuid>\`) land in the role slot; everything else falls into the picker side.
- Toggle state persists on the step config (\`recipients_custom\` / \`approvers_custom\`) so the UI re-opens in the same shape.

### Data Contract Wizard — wired-up principals
Six fields previously rendered but unwired now flow into both the submit *and* draft payloads:
| Field | Picker config | Payload key |
|---|---|---|
| Contract Owner | single, user | \`owner\` |
| Data Consumers | multi, user+group | \`consumers\` |
| Subject Matter Experts | multi, user | \`subjectMatterExperts\` |
| Read Access groups | multi, group | \`accessControl.readGroups\` |
| Write Access groups | multi, group | \`accessControl.writeGroups\` |
| Primary Support Email | single, user | \`support.primaryEmail\` |

### Backend (small, additive)
\`_resolve_role_to_users\` now comma-splits unconditionally and recurses on each segment, so the workflow step's "role + custom principals" string resolves each piece in isolation: role tokens still expand to users-in-role; emails and group names pass through as literals; dedup is enforced. The previous \`'@'\`-gated comma-split is gone; a single email still returns as one literal tuple (regression-tested).

*Note*: the plan called this a "no backend changes" phase. In practice the existing resolver only handled mixed comma lists when at least one segment contained \`@\`, which meant role + group + email wouldn't have worked end-to-end. The change here is small (one new branch at the top of the function + flatten + dedup) and strict expansion of behaviour (single-token inputs still resolve identically).

## Test plan
- [x] Backend resolver tests: 12 new in \`test_resolve_role_to_users.py\` — single-token branches still resolve (\`requester\` / \`owner\` / single email / unknown literal); comma-split mixed role+email+group; dedup; whitespace + blank segments; \`requester\` / \`owner\` inside a list still expand; legacy single-email path returns one tuple (regression).
- [x] Backend directory + resolver suite: **79 passed**.
- [x] Frontend join/split tests: 14 new in \`workflow-principals.test.ts\` — join + split round-trip identity; role-only / principals-only / mixed shapes; UUID / legacy alias / \`business:\` detection; lossless hydration; extra role-shaped tokens after the first fall into the picker side.
- [x] Frontend suite: **719 passed**, 6 skipped, 0 failed.
- [x] Type-check clean.
- [x] Lint clean on touched files (two pre-existing apostrophe errors in \`workflow-designer.tsx\` lines 1605/2016 are not part of this diff).
- [ ] Manual smoke test in a workspace — toggling the workflow designer custom-principals on/off, exercising the wizard fields end-to-end (left for reviewer / follow-up).

## Acceptance criteria from the plan (Phase 4)
- [x] Notification step toggle reveals picker; picked emails + group names appended to \`recipients\`; role/literal entries continue to work without the toggle.
- [x] Approval step toggle works for \`approvers\` with the same semantics.
- [x] Workflow execution fan-out unchanged for non-custom principals; mixed lists are resolved segment-by-segment.
- [x] Wizard owner / stakeholders / groups / primary-support-email values survive the wizard flow and appear on the submitted contract payload (and on \`Save Draft\`).
- [x] No regression to the wizard's existing summary screen or to data contracts created via the inline / basic forms.

### DIFF
--- a/src/backend/src/common/workflow_executor.py
+++ b/src/backend/src/common/workflow_executor.py
@@ -227,10 +227,33 @@ def _resolve_role_to_users(
     - '<uuid>' → app role UUID
     - 'user@email.com' → direct email
     - Legacy aliases like 'domain_owners', 'admins', etc.
+    - **Comma-separated list of any of the above** → recurse on each
+      segment, flatten results. This is what the Workflow Designer's
+      "Custom principals" toggle emits: a role/literal token from the
+      Select plus any users / groups picked from the PrincipalPicker.
     """
     from src.db_models.settings import AppRoleDb
     from src.db_models.business_roles import BusinessRoleDb
     from src.db_models.business_owners import BusinessOwnerDb
+
+    # Recursion gate for the mixed-shape list. We check for ``,`` first
+    # so each segment hits the rest of this function in isolation and
+    # role tokens / aliases / business: prefixes still resolve.
+    if ',' in role_spec:
+        out: List[tuple] = []
+        seen: set = set()
+        for segment in role_spec.split(','):
+            seg = segment.strip()
+            if not seg:
+                continue
+            for identifier, role_uuid in _resolve_role_to_users(db, seg, context):
+                if not identifier:
+                    continue
+                key = (identifier, role_uuid)
+                if key not in seen:
+                    seen.add(key)
+                    out.append((identifier, role_uuid))
+        return out
 
     # Map shorthand names to role names (legacy support)
     role_aliases = {
@@ -259,7 +282,8 @@ def _resolve_role_to_users(
         return [(role_name, role.id if role else None)]
 
     if '@' in role_spec:
-        return [(e.strip(), None) for e in role_spec.split(',')]
+        # Pre-comma-split single email -- treat as a direct recipient.
+        return [(role_spec, None)]
 
     # Business role: prefixed with "business:"
     if role_spec.startswith('business:'):
--- a/src/backend/src/tests/unit/test_resolve_role_to_users.py
+++ b/src/backend/src/tests/unit/test_resolve_role_to_users.py
@@ -1,0 +1,125 @@
+"""Unit tests for ``_resolve_role_to_users``.
+
+Pins the existing single-token behaviour and locks down the new
+comma-split / recurse path the Workflow Designer's "Custom principals"
+toggle relies on.
+"""
+
+from types import SimpleNamespace
+from unittest.mock import MagicMock
+
+import pytest
+
+from src.common.workflow_executor import _resolve_role_to_users
+
+
+def _ctx(*, user_email="alice@example.com", entity=None, entity_type=None, entity_id=None):
+    return SimpleNamespace(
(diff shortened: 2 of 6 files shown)


mvkonchits-db (approved):
LGTM — approving. Tested end-to-end locally and the stack works nicely once #424 is on top.

One thing worth confirming before this lands, and a couple of nits.

**Open question — DC wizard payload acceptance.**
This PR wires six previously-rendered-but-dropped keys into the DC create *and* update payloads: `consumers`, `subjectMatterExperts`, `accessControl.readGroups`, `accessControl.writeGroups`, `support.primaryEmail`, plus `owner` if it wasn't already plumbed. The PR description says "no backend changes" — is the `DataContractCreate` / `DataContractUpdate` Pydantic model either explicitly accepting these or set to `extra="ignore"`? If the model is strict the wizard will silently 422 on first use. A quick smoke test of a fresh DC create from the wizard after merge would lock it in.

**Nit — `KNOWN_ROLE_LITERALS` drift risk.**
`workflow-principals.ts` hardcodes the role-literal set (`requester, owner, domain_owners, project_owners, data_stewards, admins`) on the frontend. The backend equivalent lives in `workflow_executor.py` (`role_aliases`). Worth a comment on the frontend constant pointing at the backend source of truth so future role additions can't silently break the split/join round-trip.

**Nit — `_custom` toggle persisting `false` with empty principals.**
Harmless wire noise; flagging only.

**Follow-up (not blocking, applies to whole stack).**
Non-admin personas opening any surface with a `PrincipalPicker` (Comment Composer is the worst case) will see a 403 on `/api/directory/status` because the routes are gated on `settings:READ_ONLY`. The picker degrades to manual mode gracefully so there's no functional regression, but it's worth tracking — could widen `status`+`search` to a non-`settings` feature read scope or add a tiny "directory" feature with `READ_ONLY` defaulted broadly.

Nice piece of work — the abstraction held up across all 18 call sites and the fixer's pointer-events story is properly diagnosed.

Issue #281, opened 2026-04-22
Alembic multiple heads on startup: aa9_is_approver and e2_semantic_display_name


larsgeorge-db:
## Problem

On a fresh database (e.g. provisioned Lakebase), app startup crashes during `init_db` because the alembic script directory has two heads:

```
alembic.util.exc.CommandError: The script directory has multiple heads (due to branching).
Please use get_heads(), or merge the branches using alembic merge.
```

Stack trace points at `script.get_current_head()` in `src/backend/src/common/database.py:762`.

## Root cause

Both `aa9_add_is_approver_to_business_roles.py` and `a3_add_quality_items_table.py` declare `down_revision = 'z8_fix_nulls'`, creating sibling branches:

```
                 ┌── a3 → a4 → … → d1 → e2_semantic_display_name   (HEAD)
z8_fix_nulls ────┤
                 └── aa9_is_approver                                (HEAD)
```

`aa9_is_approver` was authored against the old tip before the `a3…e2` chain merged.

## Fix

Add a no-op merge revision (consistent with the existing `cfd416adf7bd` merge pattern) collapsing both heads into a single `f1_merge_aa9_e2`. No risk to existing `alembic_version` rows on either branch.

PR #420, opened 2026-05-21
build(ci): enforce single-head Alembic migration history on PRs

### LINKED ISSUE #408: [Chore]: Enforce single-head Alembic migration history on PRs
### Problem

Two PRs that branch off the same Alembic head independently produce sibling revisions, leaving `main` with multiple heads. App startup then crashes in `init_db` (`script.get_current_head()` raises *"script directory has multiple heads"*). This is exactly what happened in #281, which we resolved reactively with a manual merge revision (`f1_merge_aa9_e2`). We need a preventive CI gate so it can't recur.

Note: `src/backend/alembic.ini` does **not** pin a tip revision — Alembic's "latest" is computed from `script_location = alembic` (i.e. `src/backend/alembic/versions/`). So the actual invariant is: after merge, `alembic heads` must return exactly one head, and on a PR, that PR's new revision(s) must descend from the current `main` tip.

### Desired behavior

Add a new job `check-alembic-heads` to `.github/workflows/test-coverage.yml` (or a sibling workflow) that, on every PR:

1. Checks out the PR branch and fetches `origin/main`.
2. Computes `alembic heads` against the PR's `versions/` tree and fails if there is more than one head — unless the PR carries an `alembic-branch` label (escape hatch for the rare intentional branch + planned merge revision).
3. For each new revision file in the PR (vs. `origin/main`), asserts its `down_revision` is reachable from the head on `origin/main` — i.e. the PR was rebased onto current `main` before the migration was authored. Fails with a message pointing at `alembic merge` or rebase.
4. Auto-fix path is **out of scope** (unlike lockfiles, branching merges need human intent); just fail loudly with a clear remediation message.

### Implementation sketch

Modeled on the existing `check-lockfiles` job in `.github/workflows/test-coverage.yml` (lines 273–353):

- New job `check-alembic-heads`, `working-directory: ./src`.
- Install backend deps the same way `backend-tests` does (`pip install --require-hashes -r backend/requirements.txt`) so `alembic` is on PATH.
- Fetch base ref: `git fetch origin ${{ github.event.pull_request.base.ref }}`.
- New helper script `scripts/check-alembic-heads.sh` (or `.py`) that:
    - Runs `alembic -c backend/alembic.ini heads` and fails if more than one head is returned.
    - Diffs `backend/alembic/versions/` between PR HEAD and `origin/<base>` to find newly added revision files.
    - Loads each new revision via `alembic.script.ScriptDirectory.from_config`, reads its `down_revision`, and verifies that revision is in the ancestry of the base branch's head (computed by checking out `origin/<base>:backend/alembic/versions/` into a temp tree and running `alembic heads`).
- Skip the entire job when the PR has the `alembic-branch` label:

    ```yaml
    if: github.event_name == 'pull_request' && !contains(github.event.pull_request.labels.*.name, 'alembic-branch')
    ```

### Opt-in escape hatch

Default proposal: **PR label `alembic-branch`** — easiest to toggle, no repo changes, GH-native. Alternatives to consider at implementation time:

- Commit message trailer (e.g. `Alembic-Branch: true`) — survives squash if put in PR description.
- Marker inside the revision file itself (e.g. a non-empty `branch_labels = (...)` value or a `# alembic: branch` comment) — lives with the migration forever, not just the PR.

Implementer picks; document the choice in the PR description and in `CLAUDE.md` / contributor docs.

### Out of scope

- Auto-generating the merge revision.
- Enforcing that the merge revision (when one is intentionally created) follows a naming convention.

### References

- Past incident: #281
- Precedent CI gate shape: `check-lockfiles` job in `.github/workflows/test-coverage.yml` (lines 273–353)
- Active alembic config: `src/backend/alembic.ini`, versions in `src/backend/alembic/versions/`

will-yuponce-db:
## Summary

- Adds a new `check-alembic-heads` CI job that fails any PR which either (a) leaves the migration tree with more than one alembic head or (b) introduces a new revision whose `down_revision` is not reachable from the base branch's tip.
- Implemented in `scripts/check-alembic-heads.py`, modelled on the existing `check-lockfiles` job in `.github/workflows/test-coverage.yml:273-353`.
- Escape hatch: PRs labelled `alembic-branch` skip the gate (for the rare case where a divergent migration branch is intentional and a merge revision is planned in the same PR).
- Documented in `CONTRIBUTING.md` (new "Alembic Migration Heads" section under Supply Chain Security) and `CLAUDE.md` (new "Database Migrations" section).

Closes #408. Background incident: #281, resolved reactively with `f1_merge_aa9_e2`.

## How it works

`scripts/check-alembic-heads.py` does two things:

1. Loads `src/backend/alembic.ini` via `alembic.script.ScriptDirectory` and fails if `get_heads()` returns more than one head.
2. Uses `git diff --diff-filter=A` to find newly added `versions/*.py` files vs. `$BASE_REF`, materializes the base ref's `versions/` tree into a temp dir (via `git ls-tree` + `git show`), and verifies that every new revision's `down_revision` is either in the base set or matches another new revision in the PR (for chains of multiple revisions in one PR).

Failure paths print remediation guidance pointing at `alembic merge` or rebase, plus the bypass-label instructions.

## Local testing

- Single-head baseline (current `main`): script passes, reports `i1_workflow_version` as the head.
- Multi-head case (simulated sibling revision): script fails with annotated error and exit 1.
- Workflow YAML validated with `yaml.safe_load`.

The CI job itself will exercise both paths on this and subsequent PRs.

## Test plan

- [ ] CI runs `check-alembic-heads` on this PR and passes (no new revisions introduced here).
- [ ] Confirm the job is skipped when the `alembic-branch` label is applied (manually toggle on this PR to verify).
- [ ] Optional follow-up PR that adds a deliberate sibling revision to confirm the gate trips.

### DIFF
--- a/.github/workflows/test-coverage.yml
+++ b/.github/workflows/test-coverage.yml
@@ -352,6 +352,45 @@ jobs:
           echo "Run ./scripts/lock-requirements.sh locally and commit the updated .txt files."
           exit 1
 
+  check-alembic-heads:
+    name: Alembic Single-Head Check
+    runs-on:
+      group: databrickslabs-protected-runner-group
+      labels: linux-ubuntu-latest
+    timeout-minutes: 10
+    # Skip on PRs that intentionally carry a divergent migration branch.
+    # The label is the documented escape hatch for the rare case where two
+    # heads are deliberate and a merge revision is planned in the PR itself.
+    if: github.event_name == 'pull_request' && !contains(github.event.pull_request.labels.*.name, 'alembic-branch')
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          # Fetch enough history to diff against the PR base.
+          fetch-depth: 0
+
+      - uses: ./.github/actions/setup-python-jfrog
+        with:
+          python-version: '3.10'
+
+      - name: Install backend deps (for alembic)
+        working-directory: ./src
+        run: |
+          python -m pip install --upgrade pip
+          pip install "virtualenv<21" hatch==1.16.3
+          hatch -e dev run pip install --require-hashes -r backend/requirements.txt
+
+      - name: Fetch PR base ref
+        run: git fetch --no-tags origin "${{ github.event.pull_request.base.ref }}"
+
+      - name: Enforce single alembic head + base-ancestry
+        working-directory: ./src
+        env:
+          BASE_REF: origin/${{ github.event.pull_request.base.ref }}
+        run: |
+          hatch -e dev run python ../scripts/check-alembic-heads.py
+
   frontend-build:
     name: Frontend Build Check
     runs-on:
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -233,6 +233,20 @@ The project implements a web application designed to run as a **Databricks App**
 - Always use `hatch -e dev run ...` to run Python snippets
 - IMPORTANT: The backend server (Python) logs are in `/tmp/backend.log` and the frontend (Vite) in `/tmp/frontend.log. Read them as needed.
 
+## Database Migrations
+
+- Alembic migrations live in `src/backend/alembic/versions/`. The history must
+  always have **exactly one head**; CI enforces this via the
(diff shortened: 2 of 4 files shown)


larsgeorge-db (approved):
lgtm

PR #424, opened 2026-05-22
fix(directory): PrincipalPicker UI polish + Radix Dialog pointer-events bug

larsgeorge-db:
Stacks on **#417** (Phase 4). Post-merge polish surfaced during manual QA against the file-backed directory provider. Merge after #417.

## Root cause: Radix Dialog locks `pointer-events: none` on `<body>`

The picker's type-ahead `Popover` is portaled to `<body>` and therefore inherited that lock, making every result row unclickable inside any `Dialog` or `Sheet` host (Role form, Team form, Owner dialog, DC wizard, Workflow Designer sheet, Comment sidebar sheet, etc.). Clicks landed on the dialog's own help-text `<p>` instead.

Verified live via Playwright: `elementsFromPoint(rowCenter)` returned the dialog subtree, and ancestor walk on the popover row showed `pointer-events: none` propagating all the way from `<body>`:

```
BUTTON                     pointer-events: none
LI                         pointer-events: none
UL                         pointer-events: none
DIV (popover content)      pointer-events: none
DIV (fixed z-50 wrapper)   pointer-events: none
BODY                       pointer-events: none   ← Radix Dialog sets this
```

**Fix:** explicit ``style={{ pointerEvents: 'auto' }}`` on PrincipalPicker's `PopoverContent`. One-line change in the component covers all **18 call sites** without touching any consumer.

## Other UX issues fixed in the same pass

- **Browse sub-dialog always closes after a pick** — even in multi-select. Browse is the discovery path; the inline type-ahead is the fast path for chaining adds. Previous behaviour added the badge but left the dialog open, which was confusing.
- **Browse sub-dialog resets query + filter on re-open**, via `useEffect` on `open`. Several close paths (Close button, parent state change, single-select auto-close) bypassed the prior `onOpenChange`-based reset, so re-opening showed stale search text.
- **Inline row `onMouseDown={preventDefault}`** keeps focus on the input so the popover's dismiss-on-blur can't race the click handler. Defensive even with the pointer-events fix in place.
- **`SelectionBadges` rendered below the input row** (was inline). Pills can grow tall on multi-select; below-input layout matches Gmail-style chip lists and keeps the input width stable.
- **Dropped stale "(comma-separated)" suffix** from the *Assigned Directory Groups* label. With the directory configured, badges are discrete — the hint was misleading. A self-rendered manual-mode hint inside the picker (\"Press Enter, Tab, or comma to add an entry.\") now shows only when the directory is unconfigured. i18n updated across en, de, es, fr, it, ja, nl.

## Audit: all 18 call sites benefit

Because the fix lives inside `PrincipalPicker`, every consumer inherits it. Hosts surveyed:

| Host | Files | # |
|---|---|---:|
| Dialog | role-form, tags-settings, assign-owner, team-form, mdm/create-review, dp/team-member, dc/team-member, data-asset-reviews/create-review-request, dc/wizard (×6) | 14 |
| **Sheet** | workflows/workflow-designer (×2), comments/comment-sidebar | 3 |
| Page (no portal) | views/entitlements | 1 |

Sheet behaves identically to Dialog (both built on `@radix-ui/react-dialog`), so the same fix unbroke the workflow-designer Sheet pickers and the comment sidebar Sheet picker. The page-level entitlements call is unaffected — the fix is a no-op there.

## Tests

`principal-picker.test.tsx`: **19 passed** (was 13). New cases:
- badge-below-input order
- dialog reset-on-open
- manual-mode hint visibility (configured vs unconfigured)
- close-after-pick in multi-select
- regression: row mousedown is default-prevented

## Manual verification (Playwright-driven live browser)

1. Settings → Roles → Edit Role (Admin)
2. Type `pro` → popover opens with "Data Producers" row
3. Click row → badge added (\`[\"admins\", \"Data Producers\"]\`), input cleared ✓
4. Click *Browse directory* → type `con` → click \"Data Consumers\" row → badge added, dialog closed ✓
5. Reopen Browse after closing with `con` still in input → input is empty ✓
6. Label reads *Assigned Directory Groups* (no \"(comma-separated)\") ✓
7. Badges render below the input ✓

## Test plan

- [x] Unit: \`yarn vitest run src/components/common/principal-picker.test.tsx\` → 19/19
- [x] Manual: inline pick + Browse pick verified in live browser on Role form
- [ ] Spot-check workflow designer (Sheet host) and comment sidebar (Sheet host) — reviewer welcome to run the same scenario there

## Caveat (out of scope)

The underlying Radix pattern affects **any** portaled interactive content (custom Popover, DropdownMenu, Combobox/cmdk) rendered inside our Dialogs/Sheets. This PR fixes the directory lookup path. A broader codebase sweep for the same pattern is a follow-up.

### DIFF
--- a/src/frontend/src/components/common/principal-picker.test.tsx
+++ b/src/frontend/src/components/common/principal-picker.test.tsx
@@ -254,3 +254,128 @@ describe('PrincipalPicker — configured mode', () => {
     expect(onChange).toHaveBeenLastCalledWith(['alice@x.com']);
   });
 });
+
+describe('PrincipalPicker — manual-entry hint', () => {
+  it('renders the manual-mode hint only when directory is unconfigured', () => {
+    setDirectoryStatus(false);
+    const { rerender } = renderWithProviders(
+      <PrincipalPicker multiple value={[]} onChange={() => {}} />,
+    );
+    expect(screen.getByTestId('principal-picker-manual-hint')).toBeInTheDocument();
+
+    setDirectoryStatus(true);
+    rerender(<PrincipalPicker multiple value={[]} onChange={() => {}} />);
+    expect(screen.queryByTestId('principal-picker-manual-hint')).toBeNull();
+  });
+});
+
+describe('PrincipalPicker — pill placement', () => {
+  it('renders badges below the input row in the DOM', () => {
+    setDirectoryStatus(false);
+    renderWithProviders(
+      <PrincipalPicker multiple value={['alice@x.com']} onChange={() => {}} />,
+    );
+    const input = screen.getByTestId('principal-picker-input');
+    const badge = screen.getByTestId('principal-badge');
+    // ``compareDocumentPosition`` returns bitmask DOCUMENT_POSITION_FOLLOWING (4)
+    // when the second node follows the first. The badge should follow the input.
+    expect(input.compareDocumentPosition(badge) & Node.DOCUMENT_POSITION_FOLLOWING).toBeTruthy();
+  });
+});
+
+describe('PrincipalPicker — Browse dialog', () => {
+  beforeEach(() => {
+    setDirectoryStatus(true);
+    fetchMock.mockImplementation((url: RequestInfo | URL) => {
+      const u = url.toString();
+      if (u.includes('/api/directory/search')) {
+        return Promise.resolve(
+          new Response(
+            JSON.stringify({
+              results: [
+                { type: 'group', id: 'Producers', display_name: 'Data Producers', sub_label: 'producers-guid' },
+              ],
+            }),
+            { status: 200, headers: { 'Content-Type': 'application/json' } },
+          ),
+        );
+      }
+      return Promise.resolve(new Response('{}', { status: 200 }));
+    });
+  });
+
+  it('clears the search query and filter when the dialog is reopened', async () => {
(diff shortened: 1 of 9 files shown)


mvkonchits-db (approved):
LGTM — approving. This is the unlock; tested locally and the pickers now click inside every Dialog and Sheet host.

Verified the claims:
- `pointerEvents: 'auto'` is a single line on `PopoverContent` in `principal-picker.tsx`, unconditional, in the component itself (not duplicated at any of the 18 call sites). Right call.
- Browse sub-dialog reset moved from `handleOpenChange` (close-time, bypassed by direct `onOpenChange` callers) to `useEffect([open])` (open-time, always fires). Matches the comment about Close-button paths.
- All seven locale files (en/de/es/fr/it/ja/nl) touched in one diff for the label change.
- Row `onMouseDown e.preventDefault()` is a sensible defensive add even with the pointer-events fix in place — keeps focus on the input so the popover's dismiss-on-blur can't race the click.

Out-of-scope caveat in the PR body is fair: any *other* portaled interactive content (custom Popover/DropdownMenu/cmdk) rendered inside a Dialog or Sheet anywhere else in the codebase will hit the same Radix pattern. Tracking that broader sweep as a follow-up makes sense.

No nits.

PR #432, opened 2026-05-26
feat(workflows): preview wizard + bounded recent-executions error column

### LINKED ISSUE #405: [Feature]: Approval Wizard — preview the wizard from workflow list & details views
## Is there an existing issue for this?

## Problem statement

When designing or reviewing an approval workflow, there is currently no way to **try the wizard out** from either the workflow list or the workflow details view. The wizard only renders when a real triggering entity reaches the approval step, which makes iteration on wizard configuration painfully slow.

## Proposed Solution

Add a "Preview wizard" action to:

1. The workflow list (per-row action), and
2. The workflow details view (header action).

The preview should run the wizard against a synthetic / dry-run entity context so that designers can visually verify the wizard behaviour before deploying the workflow against real triggers. The preview must not create any audit log entries or downstream side effects.

## Additional Context

Reported as a follow-up nice-to-have ("For later") — not blocking any release. Filed so it does not get lost.

mvkonchits-db:
## Summary

Two workflow-UI improvements bundled per issue #405 review + a related polish item.

### 1. Preview wizard (closes #405)

Adds a "Preview wizard" entry point in two places:

- **Workflows list** — per-row dropdown action on approval workflows
- **Workflow designer header** — button next to the Active toggle (approval workflows only, disabled while the canvas is dirty)

Both reuse `ApprovalWizardDialog` with a new `previewMode` prop. In preview mode the dialog walks the workflow's `on_pass` graph entirely on the client — **no POST to `/api/approvals/sessions*`, no agreement persistence, no PDF generation, no delivery notifications, no audit log entries.** The side-effect-free guarantee from issue #405 is structural (no API calls happen at all), not flag-gated. Non-visual steps (`persist_agreement` / `generate_pdf` / `deliver`) show a "would …, skipped in preview" message and auto-advance via the existing chain-effect. A Preview badge in the dialog title and a dry-run hint in the description keep designers oriented.

The `on_pass` walking logic is extracted as the pure helper `computePreviewNextStep` so it can be unit-tested without driving the full Radix dialog.

### 2. Recent Executions error column — bounded rendering

The cell at `src/frontend/src/views/workflows.tsx` previously used `truncate max-w-[200px]` on an inline `<span>`. `truncate` only engages on block / inline-block elements, so multi-line stack traces stretched the row vertically and pushed the actions column off-screen. Replaced with a 2-line `line-clamp-2` button that opens a Popover containing the full message in a scrollable `<pre>` plus a copy-to-clipboard control. Row height is now stable regardless of error length.

## Files changed

- `src/frontend/src/components/workflows/approval-wizard-dialog.tsx` — `previewMode` prop, client-side step walker, preview-aware UI copy, extracted `computePreviewNextStep` helper
- `src/frontend/src/components/workflows/workflow-designer.tsx` — Preview wizard button in header
- `src/frontend/src/views/workflows.tsx` — Preview wizard row action + clamped Error column with Popover
- `src/frontend/src/components/workflows/approval-wizard-dialog.test.tsx` — 8 new unit tests for the preview walker (linear advance, non-visual passthrough, terminal detection on missing `on_pass`, terminal PASS step, non-terminal branching PASS, visual-index calculation, empty input, dangling pointers)

## Test plan

- [x] `yarn tsc --noEmit` — no new errors in changed files
- [x] `yarn test src/components/workflows --run` — 98/98 pass (was 90; +8 new)
- [x] Deployed to FEVM test app and verified manually:
  - [ ] Preview wizard launched from workflows list (row action, approval workflow only) — opens dialog with "Preview" badge, walks visual steps, shows "Preview complete" without firing onComplete
  - [ ] Preview wizard launched from workflow designer header — button hidden until workflow saved, disabled while dirty
  - [ ] No network call to `/api/approvals/sessions*` in preview (verify via browser DevTools)
  - [ ] Recent Executions row with multi-line error: error clamps to 2 lines, click opens Popover with full text + Copy works, row height stays stable

### DIFF
--- a/src/frontend/src/components/workflows/approval-wizard-dialog.test.tsx
+++ b/src/frontend/src/components/workflows/approval-wizard-dialog.test.tsx
@@ -1,0 +1,122 @@
+/**
+ * Tests for the preview-mode step walker on ApprovalWizardDialog (issue #405).
+ *
+ * The wizard's preview branch walks the workflow's on_pass graph entirely on
+ * the client — no /api/approvals/sessions* call, no agreement, no
+ * notifications. The bulk of that walker is extracted as the pure helper
+ * ``computePreviewNextStep`` so we can verify the graph rules without
+ * spinning up the Radix dialog (which is hard to drive in jsdom).
+ *
+ * What we lock down here:
+ *   1. Linear walk follows ``on_pass`` to the next step.
+ *   2. Non-visual steps (persist_agreement, generate_pdf, deliver) get
+ *      surfaced as the next step — the component's auto-advance effect
+ *      then chains them; the walker itself doesn't skip them in one tick.
+ *   3. Terminal detection: missing ``on_pass`` ⇒ complete.
+ *   4. Terminal detection: a PASS step with no outgoing edges ⇒ complete.
+ *   5. The visual-step index returned matches the ordering used by the
+ *      progress indicator (non-visual + pass/fail filtered out).
+ *   6. Empty / missing steps array is treated as terminal (safety).
+ */
+import { describe, it, expect } from 'vitest';
+import { computePreviewNextStep } from './approval-wizard-dialog';
+
+type Step = Parameters<typeof computePreviewNextStep>[0] extends (infer S)[] | undefined ? S : never;
+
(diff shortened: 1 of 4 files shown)


larsgeorge-db (approved):
lgtm

Issue #442, opened 2026-05-27
[PRD]: Unified Version Family for Data Contracts and Data Products


larsgeorge-db:
# PRD: Unified Version Family for Data Contracts and Data Products

## Problem Statement

Data Contracts and Data Products are versioned, but the version model is implemented inconsistently and only half-exposed in the UI.

1. **Two independent grouping mechanisms.** Both `data_contracts` and `data_products` carry a `parent_*_id` UUID and a `base_name` string. The version-listing query (`get_contract_versions` in `src/backend/src/controller/data_contracts_manager.py`) primarily groups by `base_name` (or worse, the mutable `name` column), and only falls back to `parent_*_id` when the string match returns empty.
2. **Inconsistent population.** The lightweight `create_new_version` path leaves both `base_name` and `parent_contract_id` `NULL` on the clone. The deep `ContractCloner` path sets `base_name` and `parent_contract_id` *and* mutates the clone's `name` to `<base>_v<version>`. The two paths leave the database in mutually incompatible states; from a clone's detail page the version family looks like a single-version family.
3. **Asymmetric UI.** Data Contracts have a version dropdown plus prev/next chevrons in the detail header. Data Products have a "New Version" button but no way to switch versions or even see siblings — and there is no `GET /api/data-products/{id}/versions` endpoint.
4. **No "latest" semantics in list views.** Both list views show every version as a separate row. Consumers see deprecated and draft rows mixed with the active ones, with no role-aware "show the version that applies to me" rule.
5. **Selectors elsewhere can't disambiguate or follow-latest.** When linking a contract to an output port, picking a product for an MDM source, etc., the dropdown shows a flat list of all versions with no scope choice. GitHub issue [#69](https://github.com/databrickslabs/ontos/issues/69) reports this concretely: same-named contracts appear identically in dropdowns. Today there is no way to express either "pin to v1.2.0" or "always follow the latest published version of this family."

Net effect: a user who clones a contract and tries to switch between versions on the detail page sees a single-version selector (it hides itself when only one row is returned), and a user picking a contract in any dropdown elsewhere in the app cannot tell duplicates apart.

## Solution

Introduce one canonical, immutable grouping key per entity — `version_family_id` (UUID) — populated consistently from every clone path. Use it as the primary lookup for "all versions of this family," with `parent_*_id` retained to encode lineage and forks.

[...]

PR #448, opened 2026-05-27
feat(versioning): role-aware family collapse, family-latest resolver, EntityVersionPicker (#442 phase 3)

### LINKED ISSUE #69: [Feature]: UI support for contract versions

### Problem statement

When I want to link a Contract directly to a Dataset, I can use the UI popup to create/edit datasets
Whenever I want to select an existing contract from the drop-down menu, I cannot see a distinction between the versions of a contract - the same contract 'name' appears twice in the list as identical values without a way to differentiate both versions.

### Proposed Solution

Add the Contract version in the drop-down menu next to the Contract name to be able to differentiate contracts

### Additional Context

_No response_

larsgeorge-db:
Phase 3 of the unified version-family work, stacked on #447. Closes #69 along the way.

## Summary

Three vertically-aligned slices, all driven by PRD #442:

1. **Role-aware visibility ranking** — new `src/common/version_visibility.py` module with elevated vs consumer status rank tables and a `collapse_by_family` helper. Wired into the contracts and products list managers, so the family representative is now picked according to caller role:
   * Consumer: `ACTIVE > DEPRECATED` (drafts hidden entirely)
   * Owner / team member / subscriber / admin: `DRAFT > PROPOSED > UNDER_REVIEW > APPROVED > ACTIVE > DEPRECATED`
   Subscription-based elevation is wired in for products via the existing `DataProductSubscriptionDb` table; contracts use ownership-only elevation (no contract subscription table exists today).

2. **Family-latest resolver endpoints** — `GET /api/data-contracts/families/{family_id}/latest` and the products counterpart. These resolve a family-follow-latest reference to a concrete row using the same role-aware rank, so any caller storing only a family id can read back the right version.

3. **EntityVersionPicker** — new shared component in `components/common/entity-version-picker.tsx`. Combobox over the collapsed list endpoint (one row per family, inline version badges), optional Entity-pinned / Family-follow-latest scope toggle, and an inline sub-picker for refining the version when pinning. The `link-contract-to-port-dialog` is migrated as the first call site (closes #69) with \`allowedScopes=['entity']\` until the output-port reference-storage slice lands.

## Closes
- #69 — same-named contracts in the link-contract-to-port dialog now show their version inline, so producers can disambiguate before linking.

## Stacked on
- #447 (list-view collapse + version-count badge)
- #446 (foundation: version_family_id column, migrations, cloning, detail-view UI)

## Test plan
- [x] \`pytest src/tests/unit/test_version_visibility.py\` — 12 cases covering status visibility, rank ordering, tie-breaks, and the consumer-only / admin-only edges. All pass.
- [x] \`pytest src/tests/unit/test_version_family.py\` — added 2 manager-level cases (consumer sees active rep; team member sees in-flight draft rep). All 17 pass.
- [ ] Manual: migrate \`link-contract-to-port-dialog\` flow — verify the picker shows version badges, that selecting a version pins the right contract id, and that the "Create New Contract" inline path still works.
- [ ] Manual: verify the family-latest resolver under both consumer and owner sessions.

## Notes
- The dev backend reloader has a pre-existing misconfiguration (called out in #447); live verification of the new endpoints will need a manual \`uvicorn\` restart. Unit-test coverage is in place.
- Family-mode writes are not yet enabled anywhere — the picker supports them, but no schema has a \`*_family_id\` reference column yet. That's the next slice and is intentionally scoped out of this PR.

## Remaining PRD scope (deferred)
- Migrate the other 5 picker call sites (the existing \`<Select>\` usages of contracts/products).
- Reference storage + indicator badge: output port → contract (family scope).
- Reference storage + badge: MDM source → contract/product.


larsgeorge-db:
Auto-closed by GitHub when stacked base branch `feat-version-family-listviews` was deleted during #447's admin merge. Reopen failed (GraphQL: Could not open the pull request — the base ref no longer exists). Re-submitted as **#467** with identical content rebased onto current `main`; #467 merged as `eb709acf`.

Issue #277, opened 2026-04-22
[FEATURE]: [UX]: Add concise guidance distinguishing Process and Approval workflows


mvkonchits-db:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Problem statement

Users see two workflow tabs ("Process workflows" / "Approval workflows") with no explanation of what each means or when to use which. The approval
  step type inside process workflows shares a name with approval workflows, causing additional confusion. New users cannot distinguish between the two
  systems without reading documentation.

### Proposed Solution

Add concise, non-intrusive guidance directly in the UI:
  1. Italic subtitles (10px) under each tab: "Background automation after events" / "Interactive consent before actions"
  2. A (?) tooltip icon next to the tabs with 3 bullet points explaining the distinction and how they chain
  3. Rename the approval step type to "Request Approval" in the designer palette, node label, and i18n to eliminate the naming collision

PR #449, opened 2026-05-27
feat(workflows): approver-role filter + webhook extras + designer fixes (Phase 1)

### LINKED ISSUE #161: [Feature]: Workflows: Filter approver roles by context-aware approval privileges
## Related

Part of PRD #77 — Complete the Process Workflow System

## Problem

The `GET /api/workflows/roles` endpoint returns **all** app roles indiscriminately. The Approvers (Role) dropdown in the workflow step configuration panel should only show roles that have the relevant `approval_privileges` flag set for the workflow's entity type(s).

For example, if a workflow triggers on `data_contract` entities, only roles with `approval_privileges.CONTRACTS = true` should appear.

A newly added "Data Governor" app role did not appear in the dropdown, exposing this gap.

## Design Decisions

- **Frontend sends `ApprovalEntity` key** as a query param to `GET /api/workflows/roles?approval_entity=CONTRACTS`. The backend filters roles by that flag.
- **Multi-entity workflows use intersection**: if a workflow trigger has multiple entity types, only roles that can approve ALL of them should be shown.
- **Entity-to-ApprovalEntity mapping** is done client-side:
  - `data_contract` → `CONTRACTS`
  - `data_product` → `PRODUCTS`
  - `domain` → `DOMAINS`
  - `data_asset_review` → `ASSET_REVIEWS`

## Backend Changes

**File:** `src/backend/src/routes/workflows_routes.py`

- Add optional `approval_entity` query param (type: `Optional[str]`) to `GET /api/workflows/roles`
- When provided, parse each role's `approval_privileges` JSON and filter to only return roles where `approval_privileges[approval_entity]` is `True`
- When not provided (backward compat), return all roles as before
- Consider accepting a comma-separated list for intersection filtering (e.g., `?approval_entity=CONTRACTS,PRODUCTS`)

## Frontend Changes

**File:** `src/frontend/src/components/workflows/workflow-designer.tsx`

- When loading roles (~line 364), map the workflow's `entityTypes` to `ApprovalEntity` keys and pass as query param
- Re-fetch roles when entity types change in the trigger configuration
- Add a constant map for the `EntityType` → `ApprovalEntity` mapping

## Acceptance Criteria

- [ ] Adding a new app role with `approval_privileges.CONTRACTS = true` makes it appear in the dropdown for contract-related workflows
- [ ] Roles without any approval privileges do not appear in the dropdown (when `approval_entity` param is provided)
- [ ] Existing behavior is preserved when no `approval_entity` param is sent (backward compatibility)
- [ ] Changing the trigger entity type in the designer refreshes the approver role list

### LINKED ISSUE #401: [Feature]: Webhook step — pass extra parameters into the UC HTTPConnection call
> Reported by external stakeholder (codename: dtag) — 2026-05-20

## Is there an existing issue for this?

## Problem statement

When a workflow Webhook step is configured to use a Unity Catalog `HTTPConnection`, we cannot connect to downstream services that require **additional parameters** beyond the connection's static configuration (e.g. extra query string params, additional headers, per-request body fields, or path segments computed from the workflow context).

Today the Webhook step passes its body template to the UC Connection call, but there is no way to provide caller-supplied parameters that the UC Connection layer would forward / merge.

## Proposed Solution

Extend the Webhook step configuration with explicit fields for parameters that are passed through to the UC `HTTPConnection` invocation:

- additional headers (key/value pairs, with template substitution)
- additional query string parameters
- path-suffix override / templating
- (TBD) per-request body augmentations beyond the existing body template

These should support `${entity.*}` and `${trigger.*}` substitution consistent with the existing body template.

## Additional Context

Related:
- #380 — enriched `entity_data` so webhook body templates can reference underlying-DP fields (e.g. `${entity.consumer_principals}`). This issue is the next layer: parameters that reach the UC Connection plumbing itself, not just the body template.

mvkonchits-db:
## Summary

Phase 1 of the open-workflow-issue follow-up. Four pieces of workflow work bundled together because they all touch the workflow authoring surface (`workflow-designer.tsx` and `workflow_executor.py`):

### 1. Filter approver roles by `approval_entity` — closes #161

`GET /api/workflows/roles` now accepts an optional `approval_entity` query param (`CONTRACTS` / `PRODUCTS` / `DOMAINS` / `ASSET_REVIEWS` / `BUSINESS_TERMS`). When set, app roles are filtered to those whose `approval_privileges[approval_entity]` is `true`. Business roles (which carry approval semantics through their own flag) always pass through.

Frontend: the Approvers (Role) picker in the workflow designer now derives the workflow's entity type(s) from the trigger config and fetches a filtered list per type, intersecting client-side so multi-entity workflows only show roles that can approve **every** entity type. Single-entity workflows make exactly one call.

No migrations — `approval_privileges` is an existing JSON column.

### 2. Webhook step extra parameters for UC HTTPConnection — closes #401

Adds three caller-supplied fields to the Webhook step config so workflows can pass context-derived headers, query string params, and path segments to downstream services beyond what the static UC `HTTPConnection` config provides:

- `additional_headers: Record<string, string>` — merged into the request headers. Caller-explicit wins on key collision.
- `additional_query_params: Record<string, string>` — appended to the URL/path as URL-encoded query string.
- `path_suffix: string` — appended to the connection's base path (UC mode) or to the URL's path component (inline mode), with single normalizing slashes.

All three fields support `${entity.*}` / `${trigger.*}` template substitution consistent with the existing `body_template`. Backward compat verified — webhook configs without these fields take the same code path as before.

UI: explicit Add / Edit / Remove UX for the two key-value fields with inline validation (empty key, duplicate key) and a non-blocking amber "Check template" badge for values with unclosed `${`. The bare `path_suffix` input has placeholder + helper text showing template substitution.

### 3. Step palette rename "Approval" → "Request Approval" — part of #277

Renames the approval step type's display label in the designer palette + node label across all 7 i18n locales (en was already correct; de/nl/fr/es/it/ja updated). Backend `step_type=approval` and DB values are untouched — purely UI-facing.

### 4. Canvas trigger node syncs with Trigger Configuration panel — pre-existing bug

Adds a small `useEffect` in the workflow designer that mirrors `triggerType` / `entityTypes` back into the canvas trigger node's `data.trigger`. Pre-existing on `main`: the Trigger Configuration panel updated the React state correctly, but the canvas node only ever displayed the `on_create` / `table` initial default and never reflected edits — surfaced during #161 testing because users couldn't visually confirm the entity-type they'd just picked.

## Files changed

**Backend:**
- `src/backend/src/controller/settings_manager.py` — `list_app_roles_for_approval(approval_entity)` (#161)
- `src/backend/src/routes/workflows_routes.py` — `approval_entity` query param on `GET /api/workflows/roles` (#161)
- `src/backend/src/common/workflow_executor.py` — `WebhookStepHandler` config additions + `_compose_path` / `_compose_url` helpers (#401)
- `src/backend/src/models/process_workflows.py` — `WebhookStepConfig` Pydantic schema additions (#401)

**Frontend:**
- `src/frontend/src/components/workflows/workflow-designer.tsx` — approver picker filter, webhook extras UI, canvas trigger node sync
- `src/frontend/src/lib/workflow-labels.ts` — `ENTITY_TYPE_TO_APPROVAL_ENTITY` constant map
- `src/frontend/src/i18n/locales/{de,es,fr,it,ja,nl}/common.json` — "Request Approval" rename
- `src/frontend/src/views/workflows.tsx` — clean tab labels (no inline subtitles; tooltip on main is unchanged)

**Tests:**
- `src/backend/src/tests/unit/test_settings_manager.py` — 4 new tests for the approval filter (no-filter, single-entity, False-exclusion, multi-entity intersection)
- `src/backend/src/tests/unit/test_webhook_step_extra_params.py` — 17 tests for additional_headers / additional_query_params / path_suffix (including inline-mode path_suffix and the legacy-config-without-extras invariant)

## Test plan

- [x] Backend unit tests pass — `28 passed` for settings manager, `17 passed` for webhook
- [x] Frontend type-check clean on touched files
- [x] E2E on a Databricks Apps workspace — created a workflow with the new webhook fields pointing at httpbin.org/post, executed it, and confirmed the echoed URL = `https://httpbin.org/post/extra/<entity_id>?caller=ontos-fevm` with `X-Trace` + `X-Phase` headers present
- [x] E2E approver-role filter — verified `?approval_entity=CONTRACTS` returns 3 app + 3 business roles vs `?approval_entity=DOMAINS` returns 2 app + 3 business roles on a workspace with realistic role data
- [x] No customer names / internal identifiers in code or commit messages

## Notes for review

- Six logical commits in this PR (one per concern + two #401 follow-ups). I considered squashing but kept them granular so the path_suffix fix and the KeyValueEditor UX rewrite are reviewable independently of the initial #401 implementation.
- The trigger-node-sync fix is a 5-line `useEffect` — happy to split into a separate small PR if you'd prefer. Bundled here because it surfaced while testing #161 and lives in the same file.

### DIFF
--- a/src/backend/src/common/workflow_executor.py
+++ b/src/backend/src/common/workflow_executor.py
@@ -1347,31 +1347,58 @@ class WebhookStepHandler(StepHandler):
         timeout_seconds = self._config.get('timeout_seconds', 30)
         success_codes = self._config.get('success_codes')
         retry_count = self._config.get('retry_count', 0)
-        
+
+        # Caller-supplied extras forwarded into the UC HTTPConnection call
+        # (issue #401). All support the same template substitution as
+        # body_template / headers; absent or empty fields are no-ops, so
+        # legacy configs are unaffected.
+        additional_headers = self._config.get('additional_headers', {}) or {}
+        additional_query_params = self._config.get('additional_query_params', {}) or {}
+        path_suffix = self._config.get('path_suffix')
+
         # Validate configuration
         if not connection_name and not url:
             return StepResult(
                 passed=False,
                 error="Webhook requires either 'connection_name' (UC Connection) or 'url' (inline mode)"
             )
-        
+
         # Substitute template variables in body
         body = None
         if body_template:
             body = self._substitute_template(body_template, context)
-        
+
         # Substitute template variables in headers
         resolved_headers = {}
         for key, value in headers.items():
             resolved_headers[key] = self._substitute_template(value, context)
-        
+
+        # Merge additional_headers (caller-supplied extras win on key collision).
+        for key, value in additional_headers.items():
+            resolved_headers[key] = self._substitute_template(value, context)
+
+        # Resolve additional_query_params (values, not keys, are templated).
+        resolved_query_params: Dict[str, str] = {}
+        for key, value in additional_query_params.items():
+            resolved_query_params[key] = self._substitute_template(value, context)
+
+        # Resolve path_suffix and merge with the configured path.
+        resolved_path_suffix = (
+            self._substitute_template(path_suffix, context) if path_suffix else ''
+        )
+        effective_path = self._compose_path(
+            base_path=path or '',
+            suffix=resolved_path_suffix,
+            query_params=resolved_query_params,
+        )
+
         try:
             if connection_name:
                 # UC Connection mode
                 result = self._execute_via_uc_connection(
                     connection_name=connection_name,
                     method=method,
-                    path=path,
+                    path=effective_path,
                     headers=resolved_headers,
                     body=body,
                     timeout_seconds=timeout_seconds,
@@ -1379,9 +1406,13 @@ class WebhookStepHandler(StepHandler):
                     retry_count=retry_count,
                 )
             else:
-                # Inline mode
+                # Inline mode — append query params to the URL.
+                effective_url = self._compose_url(
+                    base_url=url,
+                    query_params=resolved_query_params,
+                )
                 result = self._execute_direct(
-                    url=url,
+                    url=effective_url,
                     method=method,
                     headers=resolved_headers,
                     body=body,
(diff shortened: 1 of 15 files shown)


larsgeorge-db (dismissed):
lgtm

Issue #469, opened 2026-05-29
[PRD]: Ontology Term Mapping (Bulk Suggest & Apply)


larsgeorge-db:
# PRD: Ontology Term Mapping (Bulk Suggest & Apply)

## Problem Statement

Ontos already lets stewards attach customer-ontology concepts to individual entities, one at a time. On any Data Product, Data Contract (schema or property), or Asset detail page, a "Linked Terms" panel offers a single-pick concept selector that writes one row to the polymorphic semantic-link store. This is fine for occasional curation, but it does not scale to the realistic onboarding workflow:

- A steward who has just uploaded — or generated — a customer ontology covering hundreds of concepts has no way to bulk-assign those concepts across thousands of catalog assets, contract properties, and products.
- There is no AI-assisted "here are 240 columns that look like Customer Name" workflow. Every assignment is manually discovered and manually clicked.
- There is no review queue: if a suggestion engine existed, there would be nowhere for a steward to triage suggestions over multiple sessions.
- Rejected assignments are not remembered. A future suggester would re-propose the same wrong matches every time.
- There is no audit trail of *batches* of assignments, only per-link change-log entries — making it hard to answer "what did Tuesday's mapping run change?" or to roll back a single bad run.
- Onyx, a sibling project, has a working term-assignment feature with a heuristic suggester, persistent mapping, audit, diff, and rollback. We have an explicit brief to port that capability into Ontos, adapted to Ontos's data model and conventions.

Without this capability, customer ontologies in Ontos remain decorative. They live in the system but are not effectively *applied* to the catalog. The promise of ontology-driven governance (impact analysis across business terms, semantic search, automated compliance) cannot be fulfilled until the assignment layer is bulk-tractable.

## Solution

Introduce a new top-level "Term Mapping" feature under the **Govern** group: a workbench where a steward selects targets (Assets, Contract schemas / properties, Data Products) and one or more customer ontologies, runs a suggester, reviews the proposed concept assignments in a persistent queue, accepts or rejects them in bulk (with the option to override the suggested concept IRI per row), and applies the accepted set as one atomic run. Applied assignments land in the existing semantic-links store and immediately appear in the existing "Linked Terms" panels on the affected entity detail pages — no parallel storage, no new shape to read elsewhere.

[...]

PR #480, opened 2026-06-02
feat(term-mapping): backend foundation for ontology term suggester (PRD #469)

larsgeorge-db:
## Summary

Phase 1 + 2 backend for the **Term Mapping** workbench described in [PRD #469](https://github.com/databrickslabs/ontos/issues/469). Adds an end-to-end backend flow to scope a suggester run, run a deterministic heuristic engine across Assets / Data Contract schemas + properties / Data Products, persist a reviewable suggestion queue, apply accepted suggestions through the existing `SemanticLinksManager` (so the RDF graph + change log + search index all keep their familiar path), and undo on demand.

Concept candidates come strictly from **customer ontologies** (`urn:semantic-model:*`) plus opt-in shipped taxonomies; the internal `urn:taxonomy:ontos-ontology` context is permanently rejected by the manager (would otherwise let stewards map their data to Ontos's own asset-typing schema, which makes no sense).

## What's in

- **DB**: `term_mapping_runs` + `term_mapping_suggestions` tables; Alembic migration `k1_term_mapping` (single head verified).
- **Models**: SQLAlchemy `db_models/term_mappings.py` and Pydantic `models/term_mappings.py` with full `RunCreate` / `RunRead` / `SuggestionDecision` / `ApplyResult` / `UndoResult` shapes.
- **Repository**: `MappingRunRepository` + `MappingSuggestionRepository` with bulk insert and the `is_pair_already_decided` lookup the engine uses to suppress noisy re-proposals.
- **Engine package** `controller/term_mapping/`:
  - `naming.py`, `scoring.py` — pure helpers ported from `onyx_ontology` with three regression-locked bug fixes (see below).
  - `concept_source.py` — context-scoped concept loader; enforces the customer-ontology-only rule.
  - `engines/heuristic.py` — deterministic suggester (container, FK-by-stem, PK, generic property).
  - `adapters/{asset,contract,product}.py` — translate each Ontos entity family into the uniform `TargetEntity` shape.
- **Manager**: `TermMappingManager` stored on `app.state`; methods take a `Session` per request.
- **Routes**: `/api/term-mappings/*` (create/list/get runs, list/decide suggestions, apply, undo, per-entity helpers). Every endpoint gated by `PermissionChecker('term-mapping', …)`; undo is ADMIN-only.
- **Permission**: backend feature ID `term-mapping` registered in `features.py` under the Govern group.
- **Tests**: 37 new unit tests across naming, scoring, and the heuristic engine. Full unit suite still green (1063 passed, 1 skipped).

## Bug fixes inherited from porting

1. **Irregular plurals**: source's `_depluralize` only knew `-ies` / trailing-`s`, so `people` / `children` / `criteria` etc. silently dropped from matching. Added a small irregular table; tested in `test_term_mapping_naming.py`.
2. **Label-similarity cap**: source took `max(name_sim, label_sim)`, letting a long descriptive label drive an `id` column to 0.95 auto-accept against the wrong concept. Label is now a tiebreaker capped by name. Tested in `test_term_mapping_heuristic.py::test_label_similarity_capped_by_name`.
3. **Persistent rejected-pair guard**: source had no queue and re-proposed the same noise every run. The engine now consults `MappingSuggestionRepository.is_pair_already_decided` and skips known-rejected pairs. Tested in `test_term_mapping_heuristic.py::TestAlreadyDecidedGuard`.

Bonus fix found via the test suite: the self-FK guard now depluralises both sides, so a `customer_id` column on a `customers` table doesn't get a bogus FK suggestion pointing at its own concept.

## Deliberately deferred

- **Phase 3 — UI workbench** (run dialog, queue table, apply dialog, per-entity badge): separate PR. Because of that, this PR intentionally does **not** add the `term-mapping` entry to `src/frontend/src/config/features.ts` — no orphan landing-page tile leading nowhere.
- **Phase 4 — `LLMJudgeSuggester` re-ranker** + consent dialog wiring.
- Optional demo SQL seed for a retail customer ontology.

## Test plan

- [x] `hatch -e dev run pytest src/tests/unit -q` → 1063 passed, 1 skipped (locally).
- [x] `hatch -e dev run alembic heads` → single head `k1_term_mapping`.
- [x] No lints on touched files.
- [ ] Smoke-test via curl on a deployed instance:
  - `POST /api/term-mappings/runs` with `target_filter.entity_types=["asset"]` → returns `RunRead` with `status="suggested"` and a non-empty stats blob (against a workspace that has at least one customer ontology loaded).
  - `POST /api/term-mappings/runs` with `ontology_contexts=["urn:taxonomy:ontos-ontology"]` → HTTP 422 (guard fires).
  - `GET /api/term-mappings/runs/{id}/suggestions?status=pending` returns the queue with `confidence`, `reason`, `auto_apply`, `engine_metadata`.
  - `POST /runs/{id}/decisions` flips suggestion statuses.
  - `POST /runs/{id}/apply?apply_auto=true` writes `entity_semantic_links` rows; the same links surface on the affected entity's "Linked Concepts" panel.
  - `POST /runs/{id}/undo` removes exactly those links and reverts suggestion statuses.

### DIFF
--- a/src/backend/alembic/versions/k1_add_term_mapping_tables.py
+++ b/src/backend/alembic/versions/k1_add_term_mapping_tables.py
@@ -1,0 +1,93 @@
+"""add term-mapping tables
+
+Revision ID: k1_term_mapping
+Revises: j1_add_version_family_id
+Create Date: 2026-05-29 14:30:00.000000
+
+"""
+from typing import Sequence, Union
+
+from alembic import op
+import sqlalchemy as sa
+
+
+revision: str = 'k1_term_mapping'
+down_revision: Union[str, None] = 'j1_add_version_family_id'
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    op.create_table(
+        'term_mapping_runs',
+        sa.Column('id', sa.UUID(), nullable=False),
+        sa.Column('ontology_contexts', sa.JSON(), nullable=False, server_default='[]'),
+        sa.Column('include_shipped', sa.JSON(), nullable=False, server_default='[]'),
+        sa.Column('target_filter', sa.JSON(), nullable=False, server_default='{}'),
+        sa.Column('engines', sa.JSON(), nullable=False, server_default='[]'),
+        sa.Column('status', sa.String(), nullable=False, server_default='pending'),
+        sa.Column('comment', sa.Text(), nullable=True),
+        sa.Column('stats', sa.JSON(), nullable=False, server_default='{}'),
+        sa.Column('error', sa.Text(), nullable=True),
+        sa.Column('applied_link_ids', sa.JSON(), nullable=False, server_default='[]'),
+        sa.Column('created_by', sa.String(), nullable=True),
+        sa.Column('created_at', sa.TIMESTAMP(timezone=True), server_default=sa.text('now()'), nullable=False),
+        sa.Column('started_at', sa.TIMESTAMP(timezone=True), nullable=True),
+        sa.Column('finished_at', sa.TIMESTAMP(timezone=True), nullable=True),
+        sa.Column('applied_at', sa.TIMESTAMP(timezone=True), nullable=True),
+        sa.Column('undone_at', sa.TIMESTAMP(timezone=True), nullable=True),
+        sa.PrimaryKeyConstraint('id'),
+    )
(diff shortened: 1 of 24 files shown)


larsgeorge-db:
Closing in favour of a fresh implementation that uses the DataAssetReviews framework as the steward queue (MDM-style). The suggester engines, tests, and demo data fixes from this PR will be ported. See follow-up PR.

PR #481, opened 2026-06-02
feat(term-mapping): UI workbench (Phase 3)

larsgeorge-db:
## Summary

Phase 3 of the Term Mapping feature (PRD #469): surfaces the backend
suggester through a dedicated workbench under **Govern → Term Mapping**.

**Stacked on top of #480.** Base branch is `feat/term-mapping-backend`, not
`main`, so the diff shows just the UI delta. Re-target to `main` after #480
merges.

### What's new

- **\`views/term-mapping.tsx\`** — top-level workbench. Shows recent runs as a
  sortable table; clicking a run opens the detail panel with stats, the
  suggestion queue, and (when status permits) **Apply…** / **Undo** buttons.
- **\`components/term-mapping/run-config-dialog.tsx\`** — "New run" dialog:
  - Multi-select for **customer ontologies** loaded from
    \`/api/semantic-models\` (filtered to \`source_type === 'database'\` and
    \`enabled\`). Defaults to "use every enabled customer ontology" — i.e. an
    empty \`ontology_contexts\` array, which the backend interprets as the
    full set.
  - Checkbox opt-in for shipped taxonomies (Databricks ontology, ODCS).
    Note in the dialog reminds users that \`urn:taxonomy:ontos-ontology\` is
    intentionally never a valid mapping target.
  - Entity-type filter (\`asset\`, \`data_contract\`, \`data_contract_schema\`,
    \`data_contract_property\`, \`data_product\`).
  - Asset-type-names input (defaults to \`Column\`), limit, comment.
- **\`components/term-mapping/suggestion-queue-table.tsx\`** — paginated review
  table with row selection, bulk **Accept N** / **Reject N**, per-row
  custom-IRI override, confidence buckets matching the backend
  \`AUTO_ACCEPT\` / \`AUTO_REJECT\` thresholds, and an "auto" badge for
  high-confidence suggestions.
- **\`components/term-mapping/apply-dialog.tsx\`** — preview of accepted /
  auto / pending counts, \`apply_auto\` toggle, projected link-create count.
- **\`types/term-mapping.ts\`** — TS mirror of the backend Pydantic models,
  plus small UI helpers (\`bucketConfidence\`,
  \`TARGET_ENTITY_TYPE_LABELS\`, \`SHIPPED_OPT_IN_CONTEXTS\`).
- **\`App.tsx\`** route \`/term-mapping\` and **\`features.ts\`** entry under
  Govern (alpha, \`showInLanding: true\`).

### Auth surface

The view reads from \`usePermissions\` against feature ID \`term-mapping\`
(registered in #480):

- READ_WRITE gates **New run** + decide / apply.
- ADMIN gates **Undo**.
- Anyone with READ_ONLY can browse runs and the suggestion queue.

### Test plan

- [ ] Once #480 merges, restart the dev backend so the new \`term-mapping\`
      routes are live, then open http://localhost:3000/term-mapping.
- [ ] Create a run with default settings — should default to "every enabled
      customer ontology" + the heuristic engine; verify the stats card
      populates and the suggestion queue lists pending rows.
- [ ] Bulk-select a few rows → Reject; verify they move to the **Rejected**
      tab and the counts update.
- [ ] Bulk-accept → **Apply…** dialog opens; toggle \`apply_auto\` to see the
      projected count change; click Apply and verify \`links_created\` ticks
      up + the source entities gain semantic links in their detail panels.
- [ ] As ADMIN, **Undo** the run; confirm links disappear from the source
      entities and suggestions revert to pending.
- [ ] Confirm \`urn:taxonomy:ontos-ontology\` is *not* offered in either
      multi-select (the backend rejects it; UI never lists it).

### What's out of scope (follow-up tasks)

- **PendingSuggestionsBadge / "Open in Term Mapping" deep-links** on
  asset / contract / product detail panels — separate small PR once we
  agree on visual placement.
- **Phase 4 — \`LLMJudgeSuggester\`** re-ranker + LLM consent dialog wiring.
- Optional demo SQL seed for a retail customer ontology.


larsgeorge-db:
Closing in favour of a fresh implementation: move the workbench under /concepts/mapping and migrate the queue onto the existing DataAssetReviews framework (MDM-style write-through copy). See follow-up PR.

PR #505, opened 2026-06-08
feat: webhook template-variable inspector + catalog enrichment for data-product triggers

mvkonchits-db:
## Summary

- Enriches the `on_request_access` and `on_subscribe` trigger ``entity_data`` with structured catalog + output-port records so webhook ``body_template``s can reference ``\${entity.catalogs}`` and ``\${entity.output_ports}`` directly, instead of every workflow author re-parsing ODPS ``asset_identifier`` strings (`catalog.schema.table`) by hand.
- Adds a static descriptor registry on ``WorkflowsManager`` and a new ``GET /api/workflows/template-vars?trigger=<X>&entity_type=<Y>`` endpoint that returns the ``\${...}`` variables available for a given (trigger, entity_type) pair, with type, description, and sample value per descriptor.
- New ``TemplateVarsInspector`` panel in the workflow designer next to the webhook ``body_template`` field. Side panel with a search input (tokenized, matches across name + description + sample), copy-to-clipboard, collapsible groups, flat result list when searching.

## What's in scope today

Curated descriptors cover:
- ``(on_request_access, access_grant)`` — entity_data fields produced by ``AccessGrantsManager.create_request`` (request fields + the underlying DP's ``consumer_principals``, ``output_ports``, ``catalogs``)
- ``(on_subscribe, data_product)`` — entity_data fields produced by ``DataProductsManager._fire_on_subscribe_trigger`` (subscriber fields + the same DP enrichment + ``on_behalf_of`` payload)
- Universal ``flat`` group (``entity_type``, ``user_email``, ``workflow_*``, ``execution_id``) and ``context`` group (``\${context.on_behalf_of.{type,value,display}}`` for subscribe-on-behalf flows).

Other ``(trigger × entity_type)`` combinations return an empty groups list so the UI renders a friendly "no descriptors yet" state rather than 404.

## Drift mitigation

The static descriptor registry has an obvious drift risk: someone adds a new field to ``entity_data`` and forgets to update the registry. The drift-detector unit test in ``test_template_vars_registry.py`` walks every descriptor's ``\${<path>}`` through ``substitute_template`` against a fixture that mirrors the live enrichment block, and asserts no placeholder is left unrendered. New entity_data field without descriptor update → CI fails.

## Layering

All new logic lands in existing managers (``WorkflowsManager``, ``AccessGrantsManager``, ``DataProductsManager``) and the existing ``common/workflow_triggers.py`` helper module — no new feature files under ``common/``. Models live alongside the other workflow Pydantic models in ``models/process_workflows.py``. One new frontend component file (``template-vars-inspector.tsx``) since it's a UI primitive, not business logic.

## Notable side-effect

``substitute_template`` (the trigger-templating engine) is extended to JSON-encode lists/dicts and resolve one level of nested dict path (``\${entity.foo.bar}``). Required for ``\${entity.catalogs}`` (a list) to render at all — on this branch base, the engine only handled scalars. Behavior change: webhook templates that previously referenced a list/dict field would have rendered as the literal ``\${...}`` placeholder; they now render as JSON. No existing tests covered this path so there's no test churn.

## Test plan

- [x] Unit tests:
  - Catalog enrichment helper (5 cases): multi-catalog dedup, ports without parseable identifiers, caller-supplied keys preserved, empty DP, ``None`` DP
  - ``on_subscribe`` trigger enrichment mirrors the access-grant shape
  - Registry shape: every descriptor has ``path`` / ``type`` / ``description`` / ``sample`` populated
  - Registry drift detector: every documented path resolves against a realistic ``StepContext``
  - Inspector renders, copy-to-clipboard fires
- [x] FEVM end-to-end:
  - Inspector panel surfaces all three groups (entity, context, flat) on ``on_request_access × access_grant``
  - Submitted an access-grant request with ``wizard_data.on_behalf_of`` carrying ``{type: group, value: analytics_consumers}``
  - Captured the rendered webhook body via postman-echo: ``principal_type`` rendered to ``"group"``, ``principal_value`` to ``"analytics_consumers"``, ``catalogs`` to the real UC catalog array, ``output_ports`` to the structured port records


larsgeorge-db (approved):
lgtm

PR #527, opened 2026-06-18
fix(contracts): make a creator's own draft discoverable (ONT-CUJ-006)

will-yuponce-db:
## Summary

A data contract created via `POST /api/data-contracts` was saved with **no owner** — `create_contract_with_relations` never set `draft_owner_id`, and only set `owner_team_id` when an `owner` was supplied. Under the PRD #442 role-aware visibility filter, a version family is "elevated" (in-flight statuses visible) only for its `draft_owner` or `owner_team` members. So an owner-less draft was invisible to its own creator: after **Create Contract** the dialog closed and the new contract never appeared in the Data Contracts list.

Two coupled gaps:

1. **No owner on create** -> creator can't see their own draft.
2. **`GET /api/data-contracts` has no `status` query parameter**, so `?status=draft` was silently dropped by FastAPI (the documented "filter by draft" did nothing).

## Changes

- Stamp the creator as `draft_owner_id` when a draft is created without an owning team (mirrors the existing personal-draft clone path at `data_contracts_manager.py:4285`). When an owner team *is* supplied the draft stays team-visible (tier-2), so `draft_owner_id` is left unset.
- Clear `draft_owner_id` on the `draft -> proposed` transition in `request_steward_review`, promoting the contract from creator-only to organisation visibility so it's discoverable once under review.
- Add a real `status` filter to `GET /api/data-contracts`, threaded through `list_contracts_from_db` -> `_query_contracts`. It is applied **after** the visibility filter, so a caller can only ever narrow to statuses they were already permitted to see (a consumer asking `?status=draft` still gets nothing).
- Unit tests covering the `status` filter (incl. case-insensitivity) and personal-draft owner visibility (owner sees their draft; non-owners do not).

## Type of Change

Fixes regression test **ONT-CUJ-006** (Ontos CUJ regression suite — Golden Path): "After New Contract -> Create, the created contract does not appear in the Data Contracts list; the list endpoint excludes drafts and `?status=draft` is ignored."

## Testing

- `hatch -e dev run pytest src/backend/src/tests/unit/test_version_family.py` — all pass, including 4 new tests (`test_status_filter_narrows_to_single_status`, `test_status_filter_is_case_insensitive`, `test_owner_sees_own_draft_others_do_not`, `test_non_owner_consumer_sees_no_drafts`).
- Ran the contracts unit + visibility suites (`test_version_family`, `test_version_visibility`, `test_workflow_triggers_status`): 49 passed. No new failures introduced (the pre-existing route-integration failures are an unrelated "Audit service not configured" fixture gap on `main`).

## Related Issues

### DIFF
--- a/src/backend/src/controller/data_contracts_manager.py
+++ b/src/backend/src/controller/data_contracts_manager.py
@@ -2237,12 +2237,28 @@ class DataContractsManager(DeliveryMixin, SearchableAsset):
             elif not isinstance(description, dict):
                 description = {}
             
+            # A contract created without an owning team would otherwise have no
+            # owner at all, which makes it invisible to its own creator under the
+            # PRD #442 role-aware visibility filter (the family is "elevated" only
+            # for the draft_owner or owner_team members). Stamp the creator as the
+            # personal-draft owner so they can find the contract they just created.
+            # When an owner team is supplied the draft is team-visible from the
+            # start, so we leave draft_owner_id unset (tier-2 visibility). The
+            # draft_owner_id is cleared again on the draft->proposed transition.
+            status_value = data_dict.get('status', 'draft')
+            draft_owner_id = (
+                current_user
+                if (status_value or '').lower() == 'draft' and not owner_team_id
+                else None
+            )
+
             # Create main contract record
             db_obj = DataContractDb(
                 name=data_dict.get('name'),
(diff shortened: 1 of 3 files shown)


larsgeorge-db (approved):
lgtm

PR #528, opened 2026-06-18
fix(contracts): persist column-level data quality rules (ONT-CUJ-008)

will-yuponce-db:
## Summary

Quality rules authored on a schema **column** (the schema editor's per-column "Quality" tab → Add Check → name / dimension / severity / type / rule) were accepted in the dialog but **silently dropped on save**. After saving, `GET /api/data-contracts/{id}` returned `schema.quality=[]` and the column had no `quality` field, so ODCS export omitted the rules entirely.

**Root cause:** column rules arrive nested under `properties[].quality`, but `_create_schema_objects` (used by both the create and update paths) only persisted the property row itself — it never read the property's `quality` list. The only quality persistence path was the separate object-level `qualityRules` field. The `ColumnProperty` API model also didn't declare `quality`, relying on `extra="allow"` to carry it through.

## Changes

- Declare `quality: List[QualityRule]` explicitly on the `ColumnProperty` API model so the rules are first-class and survive `model_dump`.
- Extract the rule → `DataQualityCheckDb` field mapping into a shared `_build_quality_check_db(rule, object_id, property_id=None)` helper so object-level and property-level checks use identical ODCS/camelCase mapping.
- Persist `properties[].quality` in `_create_schema_objects`, bound to both `object_id` and `property_id` (so it round-trips through the existing property-level export at `data_contracts_manager.py:1063`).
- On update, scope the object-level `qualityRules` replacement to `property_id IS NULL`, so replacing object-level rules no longer deletes the column rules that the schema recreation just persisted in the same save.
- Unit tests for column-rule persistence, the no-rule case, and object/property coexistence.

## Type of Change

Fixes regression test **ONT-CUJ-008** (Ontos CUJ regression suite — Golden Path): "Column-level data-quality rules authored via the schema editor appear in the in-dialog list but are NOT persisted on Save Changes."

## Testing

- `hatch -e dev run pytest src/backend/src/tests/unit/test_contract_quality_persistence.py` — 3 new tests pass.
- `hatch -e dev run pytest src/backend/src/tests/unit/test_data_contracts_api_models.py` — all pass (the new `ColumnProperty.quality` field is non-breaking).
- No new failures across the contracts DB-integration suite (the single `test_contract_timestamps` failure is a pre-existing flake on `main`, unrelated to this change).

## Related Issues

### DIFF
--- a/src/backend/src/controller/data_contracts_manager.py
+++ b/src/backend/src/controller/data_contracts_manager.py
@@ -1652,6 +1652,7 @@ class DataContractsManager(DeliveryMixin, SearchableAsset):
         all_properties = []
         all_obj_relationships = []
         all_prop_relationships = []
+        all_quality_checks = []
         # Collect semantic link work to process after bulk insert
         schema_semantic_work = []
         prop_semantic_work = []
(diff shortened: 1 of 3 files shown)


larsgeorge-db (approved):
lgtm

PR #529, opened 2026-06-18
fix(contracts): surface proposed contracts to stewards; drop dead review stub (ONT-CUJ-012)

will-yuponce-db:
## Summary

When a producer requested a steward review, `request_steward_review` constructed an in-memory `ReviewedAsset`, logged **"Created asset review record,"** and then **never persisted it**. The result: a proposed contract was undiscoverable to stewards except by direct URL, while the logs falsely implied a review record had been created.

Proposed contracts actually reach stewards through two paths that already exist in the codebase:

1. **`GET /api/approvals/queue`** — `ApprovalsManager.get_approvals_queue` lists every contract in `proposed`/`under_review` status; the steward approves/rejects via the contract approve/reject endpoints.
2. **The `ON_REQUEST_REVIEW` workflow trigger** — runs any configured review workflow (notifications, approval routing).

The `/data-asset-reviews` surface is built for Unity Catalog assets (tables, views, functions) and `create_review_request` requires an explicit single `reviewer_email`. Routing a contract there would require a steward-assignment policy that doesn't exist in the codebase yet, so this PR intentionally does **not** invent one.

## Changes

- Remove the unsaved `ReviewedAsset` stub and its false "Created asset review record" log (and the now-unused `AssetType`/`ReviewedAssetStatus` import).
- Document the two real surfacing paths inline so the omission is intentional and clear.
- Tests: requesting review transitions `draft -> proposed` and the contract then appears in the approvals queue; requesting review from a non-draft status raises.

## Type of Change

- [x] Code cleanup / removal of misleading dead code

Addresses regression test **ONT-CUJ-012** (Ontos CUJ regression suite — Golden Path). Contract-list discoverability of proposed contracts is additionally improved by the companion draft-owner-visibility fix (#527), which clears `draft_owner_id` on the `draft -> proposed` transition and adds a `?status=` filter.

## Testing

- `hatch -e dev run pytest src/backend/src/tests/unit/test_contract_steward_review.py src/backend/src/tests/unit/test_approvals_manager.py` — 7 passed (2 new).

## Related Issues

Surfaced during manual CUJ regression testing of the Ontos app (Golden Path tab). Note: full `/data-asset-reviews` integration for contracts is deliberately out of scope pending a maintainer decision on steward/reviewer assignment.

### DIFF
--- a/src/backend/src/controller/data_contracts_manager.py
+++ b/src/backend/src/controller/data_contracts_manager.py
@@ -4847,8 +4847,7 @@ class DataContractsManager(DeliveryMixin, SearchableAsset):
         from datetime import datetime
         from src.common.workflow_triggers import get_trigger_registry
         from src.models.process_workflows import EntityType
-        from src.models.data_asset_reviews import AssetType, ReviewedAssetStatus
-        
+
         contract = data_contract_repo.get(db, id=contract_id)
         if not contract:
             raise ValueError("Contract not found")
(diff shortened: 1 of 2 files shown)


larsgeorge-db (approved):
lgtm

PR #530, opened 2026-06-18
fix(contracts): allow approving a contract directly from proposed (ONT-CUJ-013)

will-yuponce-db:
## Summary

`POST /api/data-contracts/{id}/approve` returned **HTTP 500** (`"Failed to approve contract"`) for a `proposed` contract, leaving it stuck in `proposed` with no error toast in the UI.

**Root cause:** the approve endpoint is documented as `PROPOSED/UNDER_REVIEW -> APPROVED` and explicitly pre-checks that the source status is one of those — but `DATA_CONTRACT_TRANSITIONS["proposed"]` only listed `["draft", "under_review", "deprecated"]`. `"approved"` was missing, so `transition_status(proposed -> approved)` raised `ValueError("Invalid status transition")`, which the route swallowed with a bare `except Exception` and re-raised as an opaque 500.

## Changes

- Add `"approved"` to the allowed transitions from `"proposed"` in `DATA_CONTRACT_TRANSITIONS`, so a steward can approve directly. `under_review` remains an optional intermediate step rather than a mandatory one — consistent with the route's documented behavior.
- In the `approve` and `reject` routes, catch `ValueError` and return **409** with the message instead of an opaque 500, so any genuinely invalid transition is a clean client error.
- Tests: the lifecycle map allows `proposed -> approved`; `transition_status` succeeds for `proposed -> approved`; a genuinely invalid jump (`active -> approved`) still raises.

## Type of Change

Fixes regression test **ONT-CUJ-013** (Ontos CUJ regression suite — Golden Path): "Steward 'Approve' on the proposed contract fails with HTTP 500; contract remains 'proposed'."

Note on separation-of-duties: the regression note also observed that a self-approval (approver == author) should ideally return a clean 403. The codebase has no SoD check today, so the 500 was purely the transition-map bug fixed here; an explicit author≠approver guard is left as a separate enhancement pending a maintainer decision.

## Testing

- `hatch -e dev run pytest src/backend/src/tests/unit/test_workflow_triggers_status.py` — 15 passed (3 new).

## Related Issues

### DIFF
--- a/src/backend/src/models/lifecycle.py
+++ b/src/backend/src/models/lifecycle.py
@@ -48,7 +48,10 @@ DATA_PRODUCT_TRANSITIONS: dict[str, list[str]] = {
 
 DATA_CONTRACT_TRANSITIONS: dict[str, list[str]] = {
     "draft": ["proposed", "deprecated"],
-    "proposed": ["draft", "under_review", "deprecated"],
+    # A steward can approve directly from "proposed" (the approve endpoint is
+    # documented as PROPOSED/UNDER_REVIEW -> APPROVED). "under_review" stays an
+    # optional intermediate step rather than a mandatory one.
+    "proposed": ["draft", "under_review", "approved", "deprecated"],
     "under_review": ["draft", "approved", "deprecated"],
     "approved": ["active", "draft", "deprecated"],
     "active": ["deprecated"],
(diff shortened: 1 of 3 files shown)


larsgeorge-db (approved):
lgtm

Issue #347, opened 2026-05-07
[Feature]: Scope Data Consumer asset visibility to assets linked from accessible Data Products


mvkonchits-db:
## Context

A customer using Ontos with role-based access flagged a visibility gap in how Assets are exposed to Data Consumers.

## Current behavior

Two failure modes depending on whether the Data Consumer role has the `assets` feature granted:

1. **`assets` feature granted (READ_ONLY):** Data Consumers see *all* assets in the Asset Explorer, regardless of whether those assets are linked to a Data Product they can access. Over-shares.
2. **`assets` feature not granted:** Data Consumers cannot see linked assets *inside* a Data Product detail view either — the `Linked Assets` panel falls back to empty because the asset fetch is gated by the same feature flag. Under-shares.

Neither mode matches intent.

## Desired behavior

A Data Consumer should see assets *scoped to the Data Products they can access*:

- In the **DP detail view's `Linked Assets` panel** — show assets linked to *this* DP, regardless of `assets` feature state.
- In the **Asset Explorer** (if exposed to consumers at all) — show only assets linked to at least one DP the consumer can access.

## Why this matters

- Without the feature flag, consumers can't browse what they're entitled to.
- With the feature flag, consumers see assets they have no business seeing — a least-privilege violation in deployments that take role boundaries seriously.

## Implementation considerations (not prescriptive)

Two candidate paths:

- **Frontend filter** — smaller change, but leaks data over the wire and through direct API calls.
- **Backend scoped endpoint / role-aware filter** — add DP-scoping to `GET /api/assets` (or introduce a `GET /api/data-products/{id}/assets` endpoint). The consumer's accessible-DP set drives the asset query. Larger change, secure by default.

Backend route is the safer of the two; worth a quick scoping discussion before implementation.

## Out of scope

- Ontology changes
- Permission-model changes for non-Consumer roles

## Acceptance criteria

- [ ] Data Consumers see `Linked Assets` in DP detail view even without `assets` feature granted
- [ ] Data Consumers do not see assets unrelated to DPs they can access, regardless of `assets` feature state
- [ ] Existing behavior for Data Producers / Admins unchanged
- [ ] Tests cover both the linked-from-DP path and the broader Asset Explorer path

mvkonchits-db:
PR opened: #349 — backend-scoping route, 18 unit tests, real-workspace E2E verified both acceptance criteria. Closing via that PR on merge.

PR #531, opened 2026-06-18
fix(data-products): make products editable by creator and feature admins (ONT-CUJ-015)

will-yuponce-db:
## Summary

Adding a deliverable to a data product failed with `PUT /api/data-products/{id}` → **403** for both the Producer who created it *and* for an Admin. Three coupled root causes:

1. **`create_product` never assigned an owner.** A new product had no `project_id`, `owner_team_id`, or `draft_owner_id`, so the update authorization cascade failed for everyone except a workspace admin. The creator couldn't edit their own product — and it could never be given an owner, since assigning one also requires `PUT`.
2. **The update ownership cascade only short-circuited on workspace admin** (group-based `is_user_admin`), ignoring the caller's **feature-level** `data-products` permission. An in-app role override granting `data-products=Admin` was therefore denied when the caller wasn't the object owner (the "Admin retry still 403" finding).
3. **`GET /api/delivery-methods` required the `delivery-methods` feature**, which a Producer lacks — so the Add Deliverable delivery-method dropdown was empty (403).

## Changes

- `create_product` stamps the creator as `draft_owner_id` when no project, team, or draft owner is supplied (mirrors the personal-draft clone path). Team-owned products are left team-visible.
- `update_product_with_auth` accepts `is_feature_admin` and short-circuits on it in addition to workspace admin. The update route resolves the caller's effective `data-products` permission — honoring in-app role overrides via the same `applied_role_override → effective-permissions → ADMIN` resolution the DP-assets route uses — and passes it in.
- Gate the delivery-methods **read** endpoints (`GET ""`, `GET /{id}`) on `data-products:READ_ONLY` — reading the catalog is a prerequisite for authoring deliverables (same precedent as the DP linked-assets route, Issue #347). **Write** endpoints stay on `delivery-methods:READ_WRITE` (admin-managed catalog).
- Tests: creator becomes draft owner on create; an explicit owner team is not overwritten; a feature-admin can edit an orphan and others' products; non-feature-admins remain fail-closed.

## Type of Change

Fixes regression test **ONT-CUJ-015** (Ontos CUJ regression suite — Golden Path): "Add Deliverable submit -> 403 (deliverable not saved); delivery-methods 403 for Producer; product update gated by object-level ownership that ignores the Admin feature permission, and new products saved with no owner -> uneditable by anyone."

## Testing

- `hatch -e dev run pytest src/backend/src/tests/unit/test_data_products_ownership_scope.py` — 28 passed (5 new).
- Broader product manager/route suites show no new failures vs. `main` (pre-existing failures are an unrelated "Audit service not configured" fixture gap).

## Related Issues

Surfaced during manual CUJ regression testing of the Ontos app (Golden Path tab). This is the data-products analogue of the contract draft-ownership fix (#527).

### DIFF
--- a/src/backend/src/controller/data_products_manager.py
+++ b/src/backend/src/controller/data_products_manager.py
@@ -164,6 +164,21 @@ class DataProductsManager(DeliveryMixin, SearchableAsset):
             product_data.setdefault('kind', 'DataProduct')
             product_data.setdefault('status', DataProductStatus.DRAFT.value)
 
+            # Stamp the creator as the single-user owner when no other owner is
+            # supplied. Without this, a freshly created product has no
+            # project_id, owner_team_id, or draft_owner_id, so the update
+            # authorization cascade fails for *everyone* except a workspace
+            # admin — the creator can't even add deliverables to their own
+            # product (ONT-CUJ-015). Mirrors the personal-draft clone path.
+            if (
+                user
+                and not product_data.get('draft_owner_id')
+                and not product_data.get('draftOwnerId')
+                and not product_data.get('owner_team_id')
+                and not product_data.get('project_id')
+            ):
+                product_data['draft_owner_id'] = user
+
             # Validate
             try:
                 product_api_model = DataProductCreate(**product_data)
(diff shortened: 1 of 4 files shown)


larsgeorge-db (approved):
lgtm

PR #533, opened 2026-06-18
fix(contracts): reject duplicate contract name within the same domain (ONT-NEG-002)

will-yuponce-db:
## Summary

Creating a data contract whose name duplicated an existing contract in the same domain **succeeded silently** — `POST /api/data-contracts` returned 200 with a duplicate persisted, no 409 and no inline validation error.

`create_contract_with_relations` always creates a **new version family** (new versions of an existing contract go through `create_version`), so any existing contract sharing the same `name` + `domain` is a genuine duplicate, not another version of the same family.

## Changes

- Add a case-insensitive `name` + `domain_id` duplicate check in `create_contract_with_relations`, raising `ConflictError` (HTTP 409) — consistent with the existing duplicate-id behavior for data products and the `ConflictError` pattern used by the projects and delivery-methods managers. The check runs before any writes.
- A contract with the same name in a **different** domain is still allowed; versioning (same name within a family) is unaffected since it uses a different code path.
- Unit tests: duplicate name in the same domain raises `ConflictError` (including case-insensitive match); a distinct name still creates successfully.

## Type of Change

Fixes regression test **ONT-NEG-002** (Ontos CUJ regression suite — Negative Paths): "Creating a contract named 'Customer Data Contract' (duplicate of an existing active contract of the same name/domain) succeeded; expected 409/inline error and no duplicate."

## Testing

- `hatch -e dev run pytest src/backend/src/tests/unit/test_contract_duplicate_name.py` — 3 passed.
- Contract create / version-family suites show no new failures (the single `test_contract_timestamps` failure is a pre-existing flake on `main`).

## Related Issues

### DIFF
--- a/src/backend/src/controller/data_contracts_manager.py
+++ b/src/backend/src/controller/data_contracts_manager.py
@@ -2211,7 +2211,29 @@ class DataContractsManager(DeliveryMixin, SearchableAsset):
                 domain_id = self._resolve_domain(db, domain_id=data_dict.get('domainId'))
             elif data_dict.get('domain'):
                 domain_id = self._resolve_domain(db, domain_name=data_dict.get('domain'))
-            
+
+            # Reject a duplicate contract name within the same domain. This path
+            # always creates a *new* version family (new versions go through
+            # create_version), so any existing contract sharing the name+domain
+            # is a genuine duplicate, not another version of the same family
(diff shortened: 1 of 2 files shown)


larsgeorge-db (approved):
lgtm

PR #534, opened 2026-06-18
fix(contracts): require a schema before proposing a contract for review (ONT-NEG-005)

will-yuponce-db:
## Summary

An empty draft contract (no schema objects) could be proposed for review: `POST /api/data-contracts/{id}/request-review` returned 200 with `status -> proposed`. There is nothing for a steward to review on an empty draft, and the documented "schema required before proposal" guard was never enforced.

## Changes

- `request_steward_review` raises `ValueError` when the contract has zero schema objects, leaving the contract in `draft`.
- The request-review route now surfaces the validation message on the 409 response (previously masked as a generic "Invalid review request") so the UI can explain *why* the request was blocked. A "not found" error still maps to 404.
- Unit tests: an empty draft cannot be proposed (and stays `draft`); a draft with a schema object can.

## Type of Change

Fixes regression test **ONT-NEG-005** (Ontos CUJ regression suite — Negative Paths): "A Draft contract with no schema objects can be proposed for review; the documented 'schema required before proposal' guard is not enforced."

## Testing

- `hatch -e dev run pytest src/backend/src/tests/unit/test_contract_review_schema_guard.py` — 2 passed.

## Related Issues

### DIFF
--- a/src/backend/src/controller/data_contracts_manager.py
+++ b/src/backend/src/controller/data_contracts_manager.py
@@ -4919,7 +4919,19 @@ class DataContractsManager(DeliveryMixin, SearchableAsset):
         from_status = (contract.status or '').lower()
         if from_status != 'draft':
             raise ValueError(f"Cannot request review from status {contract.status}. Must be DRAFT.")
-        
+
+        # A contract must describe at least one schema object before it can be
+        # proposed for review — there is nothing for a steward to review on an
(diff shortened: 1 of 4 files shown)


larsgeorge-db (dismissed):
lgtm

larsgeorge-db (approved):
lgtm

PR #535, opened 2026-06-18
fix(data-products): gate direct reads of unpublished products (ONT-NEG-011)

will-yuponce-db:
## Summary

`GET /api/data-products/{id}` returned the full product to any caller with `data-products:READ_ONLY`, with **no status or ownership check**. A Data Consumer could read a **draft** (uncertified) product directly by id — `GET /api/data-products/{draftId}` → 200 with the full draft — even though the marketplace listing correctly hides drafts.

## Changes

A read gate on the get-by-id route (`_caller_can_read_product`):

- **Published** products (`active`/`deprecated`) remain readable by anyone with `READ_ONLY` — the catalog/marketplace contract.
- **Unpublished** products (`draft`/`proposed`/`under_review`/`approved`/...) are readable only by data-products **admins** (including via an in-app role override) and **owners** (`draft_owner` / owning team / project member) — reusing the same ownership scope the listing uses (`list_products` + role-override resolution), consistent with the DP-assets route.
- A denied read returns **404** (not 403) so the draft's existence isn't disclosed.

## Type of Change

Fixes regression test **ONT-NEG-011** (Ontos CUJ regression suite — Negative Paths): "A Consumer can directly read a DRAFT (uncertified) product via the API; expected 404/not-available. (The marketplace listing already hides drafts, but direct API access was not gated.)"

## Testing

- `hatch -e dev run pytest src/backend/src/tests/unit/test_data_product_read_gate.py` — 4 passed. The gate is tested directly (published readable by all; draft denied for a non-owner consumer; draft allowed for a feature admin and for an owner in the accessible set).

## Related Issues

### DIFF
--- a/src/backend/src/routes/data_product_routes.py
+++ b/src/backend/src/routes/data_product_routes.py
@@ -64,6 +64,81 @@ def get_data_products_manager(
     return manager
 
 
+def _caller_can_read_product(request, db, current_user, manager, product) -> bool:
+    """Whether ``current_user`` may read ``product`` directly.
+
+    Published products (active/deprecated) are readable by any caller with
+    data-products READ_ONLY — that's the catalog/marketplace contract.
+    Unpublished products (draft/proposed/under_review/approved/...) are only
+    readable by data-products admins (incl. via an in-app role override) and
+    by owners (draft_owner / owning team / project member). This stops a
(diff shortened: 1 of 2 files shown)


larsgeorge-db (approved):
lgtm

PR #536, opened 2026-06-19
fix(semantic-models): switch concept_iri routes to ?iri= to survive %2F%2F path collapse

mvkonchits-db:
## Summary

Clicking any concept detail page (`/concepts/browser/{iri}`) renders "Concept not found", regardless of which concept or ontology. Blocks the entire Concepts CUJ (ONT-FEAT-001 in regression sheet).

**Root cause:** the Databricks Apps proxy issues a 301 that decodes and collapses `%2F%2F` → single `/` in path segments, mangling IRIs like `http://ontos.example.org/…` into `http:/ontos.example.org/…` before they reach FastAPI. `%23` and `%3A` are preserved; only the double-slash collapses. RDFLib then fails the lookup → 404.

Query-string values survive the transformation unchanged. Sibling endpoints in the same file (`/concepts/hierarchy?iri=`, `/neighbors?iri=`) already use the query-param shape and work fine — this PR aligns the rest of the `{concept_iri:path}` route family on the same pattern.

## Approach

- Add new `/concepts/by-iri?iri=<urlencoded>` (semantic-models) and `/knowledge/concepts/by-iri?iri=<urlencoded>` (knowledge) variants for every endpoint that used `{concept_iri:path}` — ~14 routes total.
- The path-form `/concepts/{concept_iri:path}` is **kept as a deprecated alias** so external bookmarks/integrations don't 404 during the migration window. `deprecated=True` is set on the FastAPI decorator so OpenAPI/Swagger flags them.
- Both shapes delegate to shared module-level `_*_payload` helpers — no duplicated logic; either route family stays in sync.
- Frontend (`concept-detail.tsx`, `node-links-panel.tsx`, `concepts-search.tsx`) switched to the query-param form.
- Drive-by fix: `get_concept_hierarchy` referenced an undefined name `concept_iri` inside its exception logger (parameter is named `iri`).

### Why `/by-iri` and not bare `/concepts?iri=`

`GET /api/semantic-models/concepts` already serves the list endpoint (`?q=&limit=…`). Adding a bare `?iri=` form would collide with the list endpoint signature. The path-prefix tweak keeps the new shape explicit.

### Why route ordering matters

`/by-iri` routes are declared **before** the `{concept_iri:path}` ones. FastAPI matches in declaration order and the path parameter is greedy, so the reverse order would shadow the new routes. A comment in the source explains this.

## Test plan

- [x] Unit tests for both route shapes (`tests/integration/test_concept_iri_routing.py`): query-param 200, deprecated path-form 200, PATCH/DELETE on both.
- [x] `yarn type-check` passes (no new errors)
- [ ] Manual: deploy + click any concept on `/concepts` — detail panel renders (not "Concept not found")
- [ ] Manual: hit each deprecated path-form route; confirm 200 and FastAPI deprecation marker shows in docs

## Migration

Deprecated path-form routes can be removed after callers migrate. **No alembic migrations required.**

### DIFF
--- a/src/backend/src/routes/semantic_models_routes.py
+++ b/src/backend/src/routes/semantic_models_routes.py
@@ -630,35 +630,71 @@ async def get_concept_hierarchy(
     try:
         logger.info(f"Retrieving hierarchy for concept: {iri}")
         hierarchy = manager.get_concept_hierarchy(iri)
-        
+
         if not hierarchy:
             raise HTTPException(status_code=404, detail="Concept not found")
-        
+
         return {'hierarchy': hierarchy.model_dump()}
     except HTTPException:
         raise
     except Exception as e:
-        logger.error("Error retrieving concept hierarchy for %s", concept_iri, exc_info=True)
+        logger.error("Error retrieving concept hierarchy for %s", iri, exc_info=True)
         raise HTTPException(status_code=500, detail="Failed to retrieve concept hierarchy")
 
-@router.get('/semantic-models/concepts/{concept_iri:path}')
+# -------- /semantic-models/concepts detail (by IRI) --------
+#
+# The path-form route (``/concepts/{concept_iri:path}``) is preserved as a
+# deprecated alias so existing bookmarks keep working, but the canonical form
+# uses a query parameter. This avoids a proxy bug where some HTTP intermediaries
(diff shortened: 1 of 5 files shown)


larsgeorge-db (approved):
lgtm

PR #537, opened 2026-06-19
fix(compliance): drop id requirement from policy create body (#235)

### LINKED ISSUE #235: [BUG]: Cannot create new compliance policies from the UI — 422 "field required" on missing id

### Current Behavior

  Current Behavior                                                                                                                                      
                                                                                                                                                      
  Creating a new compliance policy from the UI (Compliance → Create Policy) always fails with a 422 Unprocessable Entity error, regardless of what
  values are entered in the form. The error is "field required" for the id field.                                                                       
   
  Editing, viewing, running, and deleting existing policies (e.g., from demo data) all work correctly. Only creation is broken.                         
                                                                                                                                                      
  Root Cause                                                                                                                                            
                                                                                                                                                      
  The POST /api/compliance/policies route expects a CompliancePolicy Pydantic model as the request body (compliance_routes.py:61):                      
   
  policy: CompliancePolicy = Body(...)                                                                                                                  
                                                                                                                                                        
  The CompliancePolicy model (models/compliance.py:8-9) defines id as required with no default:                                                         
                                                                                                                                                        
  class CompliancePolicy(BaseModel):                                                                                                                    
      id: UUID = Field(...)  # required, no default, no auto-generation                                                                               

  The frontend (compliance.tsx:143-150) correctly omits id when creating a new policy (there is no ID yet):                                             
   
  const createPayload: Omit<CompliancePolicy, 'id'> = {                                                                                                 
    ...policyData,                                                                                                                                    
    compliance: 0,                                                                                                                                      
    history: [],
    created_at: new Date().toISOString(),                                                                                                               
    updated_at: new Date().toISOString(),                                                                                                               
  };
                                                                                                                                                        
  Pydantic validation rejects the payload before the route handler executes because id is missing. 

### Expected Behavior

New compliance policies should be created successfully. The id should be auto-generated server-side.

**Suggested Fix**                                                                                                                                         
                                                                                                                                                      
  Option A (preferred): Create a separate CompliancePolicyCreate model for the POST route that excludes id:                                             
  
  class CompliancePolicyCreate(BaseModel):                                                                                                              
      name: str = Field(...)                                                                                                                          
      description: str = Field(...)                                                                                                                     
      failure_message: Optional[str] = None
      rule: str = Field(...)                                                                                                                            
      is_active: bool = True                                                                                                                            
      severity: str = "medium"
      category: str = "general"                                                                                                                         
                                                                                                                                                      
  Option B: Make id optional with auto-generation:                                                                                                      
  
  id: UUID = Field(default_factory=uuid4)                                                                                                               
                                                                                                                                                      
  Note: the compliance field also has this pattern (compliance: float = Field(...) with no default), but the frontend works around it by hardcoding     
  compliance: 0 in the create payload. Adding default=0.0 would be a good defensive fix.

### Steps To Reproduce

1. Navigate to Compliance
  2. Click "Create Policy"
  3. Fill in all form fields (name, description, rule, category, severity)                                                                              
  4. Click Save
  5. Observe 422 error / "Failed to save policy" toast 

### Cloud

AWS

### Browser

Chrome

### Version

latest

### Relevant log output

```shell

```

mvkonchits-db:
Closes #235

## Summary

Creating a new compliance policy from the UI (Compliance → Create Policy) always failed with `422 Unprocessable Entity` / "Error: Field required" regardless of form contents. Blocks the entire Compliance CUJ (ONT-FEAT-010 P1 in the regression sheet).

## Root cause

The POST `/api/compliance/policies` route bound `policy: CompliancePolicy = Body(...)`. `CompliancePolicy` was the **read** model and marked `id: UUID = Field(...)` and `compliance: float = Field(...)` as required (no defaults). The frontend correctly omits `id` on create — there is no UUID yet — so Pydantic rejects the body before the handler runs.

Classic shared-CRUD-model anti-pattern. Same model used for both POST (create) and PUT (update) where the contracts genuinely differ.

## Fix

- Added `CompliancePolicyCreate` (in `models/compliance.py`): only `name`, `description`, `failure_message`, `rule`, `is_active`, `severity`, `category`. No `id`, no `compliance`, no `history`, no timestamps — all server-generated.
- POST `/api/compliance/policies` now binds `policy: CompliancePolicyCreate = Body(...)`.
- `compliance_manager.create_policy` accepts `CompliancePolicyCreate` and generates `id = str(uuid.uuid4())` server-side. The DB schema (`compliance_policies.id = Column(String, primary_key=True)`) already has no DB-level default, so this is the natural place for it.
- Defensive secondary fix (per the issue body): `CompliancePolicy.compliance` now has `default=0.0` so older rows / future contexts don't 500 on reads.
- PUT route and `update_policy` manager are **unchanged** — they still use the full `CompliancePolicy` model, which is correct for a full-document update where the `id` comes from the URL path.

## No alembic migration required

Verified explicitly: the DB schema is unchanged. `compliance_policies.id` is already `Column(String, primary_key=True)` with no DB default; the manager always supplies the UUID. `created_at`/`updated_at` already have SQLAlchemy-level defaults. The fix is purely Pydantic + manager wiring.

## Behavior with client-provided id

If an old client still sends `id` in the create body, the new model **ignores it** (extra fields are dropped by Pydantic v2 default). Locked in by `test_create_policy_ignores_client_provided_id`. No 422 for legacy clients.

## Test plan

- [x] `hatch -e dev run pytest src/backend/src/tests/unit/test_compliance_manager.py` — 21/21 pass
- [x] `hatch -e dev run pytest src/backend/src/tests/integration/test_compliance_routes.py::test_create_policy_success` — passes
- [x] New: `test_create_policy_ignores_client_provided_id` — passes
- [x] New: `test_create_policy_generates_unique_ids` — passes (replaces the old `test_create_policy_duplicate_id` which no longer applies — duplicates can't come through the create contract)
- [ ] Manual: deploy + open `/compliance` → Create Policy → fill `name`, `description`, `rule`, `category`, `severity` → Save → policy appears in list with status Active

### Pre-existing test infra gap (out of scope, noted for transparency)

`test_get_policies`, `test_update_policy`, `test_get_policy_by_id` fail on `upstream/main` HEAD as well — the integration `client` fixture in `conftest.py` does not wire `AuditManager` into `app.state` and some assertions expect a dict shape that the route now returns as a list. **Not regressions from this PR.** Worth a separate cleanup PR.

### DIFF
--- a/src/backend/src/controller/compliance_manager.py
+++ b/src/backend/src/controller/compliance_manager.py
@@ -13,7 +13,7 @@ from src.repositories.compliance_repository import (
     compliance_run_repo,
     compliance_result_repo,
 )
-from ..models.compliance import CompliancePolicy, ComplianceRun, ComplianceResult
+from ..models.compliance import CompliancePolicy, CompliancePolicyCreate, ComplianceRun, ComplianceResult
 from src.common.compliance_dsl import evaluate_rule_on_object as eval_dsl, parse_rule, Evaluator
 from src.common.compliance_actions import ActionExecutor, ActionContext, get_action_registry
 from src.common.compliance_entities import (
@@ -168,13 +168,13 @@ class ComplianceManager:
         return db.get(CompliancePolicyDb, policy_id)
 
     def create_policy(
-        self, 
-        db: Session, 
-        policy: CompliancePolicy,
+        self,
+        db: Session,
+        policy: CompliancePolicyCreate,
         current_user: Optional[str] = None,
     ) -> CompliancePolicyDb:
         db_obj = CompliancePolicyDb(
-            id=str(policy.id),  # Convert UUID to string for SQLite
+            id=str(uuid.uuid4()),  # Server-generated UUID
             name=policy.name,
             description=policy.description,
             failure_message=policy.failure_message,
--- a/src/backend/src/models/compliance.py
+++ b/src/backend/src/models/compliance.py
@@ -5,13 +5,24 @@ from uuid import UUID
 from pydantic import BaseModel, Field
 
 
+class CompliancePolicyCreate(BaseModel):
+    """Payload accepted by POST /api/compliance/policies. id, compliance score, and timestamps are server-generated."""
+    name: str = Field(..., description="Name of the compliance policy")
+    description: str = Field(..., description="Description of what the policy checks")
+    failure_message: Optional[str] = Field(default=None, description="Human-readable message shown when policy fails")
+    rule: str = Field(..., description="Policy rule using the compliance DSL")
+    is_active: bool = Field(default=True, description="Whether the policy is active")
+    severity: str = Field(default="medium", description="Severity level: low, medium, high")
+    category: str = Field(default="general", description="Category of the policy")
+
+
 class CompliancePolicy(BaseModel):
     id: UUID = Field(..., description="Unique identifier for the policy (UUID)")
     name: str = Field(..., description="Name of the compliance policy")
     description: str = Field(..., description="Description of what the policy checks")
     failure_message: Optional[str] = Field(default=None, description="Human-readable message shown when policy fails")
     rule: str = Field(..., description="Policy rule using the compliance DSL")
-    compliance: float = Field(..., description="Current compliance score (0-100)")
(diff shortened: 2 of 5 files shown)


larsgeorge-db (approved):
lgtm

PR #540, opened 2026-06-19
fix(ontology): persist concept_type and property metadata in RDF

larsgeorge-db:
## Summary

Custom Properties created via the concept editor were always rendered as plain Concepts in the UI. The backend was hardcoding `skos:Concept` on write and dropping the property-specific fields (`property_type`, `domain`, `range`), so there was no way to recover the user's intent on read.

This PR wires those fields end-to-end so every option in the editor dropdown round-trips correctly:

- `concept` (default) → `skos:Concept`
- `class` → `owl:Class` + `rdfs:Class`
- `property` + `object` → `owl:ObjectProperty` + `rdf:Property` (+ optional `rdfs:domain`/`rdfs:range`)
- `property` + `datatype` → `owl:DatatypeProperty` + `rdf:Property`
- `property` + `annotation` → `owl:AnnotationProperty` + `rdf:Property`
- `individual` → `owl:NamedIndividual`
- `term` → `skos:Concept` + custom `ontos:conceptType "term"` annotation so the user-facing type can be reconstructed

### Changes

- `models/ontology.py` — expose `property_type` / `domain` / `range` on `OntologyConcept`, `ConceptCreate`, `ConceptUpdate`, plus `concept_type` on update for draft retypes.
- `controller/semantic_models_manager.py`
  - New `_resolve_concept_rdf_types` and `_detect_concept_type` helpers centralise the mapping in both directions.
  - `create_concept` now emits the resolved `rdf:type` set and writes `rdfs:domain` / `rdfs:range` for properties.
  - `update_concept` supports draft retyping and partial `domain`/`range` updates without clobbering unspecified fields.
  - Read paths (`_extract_concept_metadata`, `get_concept`, `get_concept_details`, `_compute_all_concepts`) surface `property_type` / `domain` / `range`.
  - `promote_concept` / `migrate_concept` forward the new fields.
- `routes/semantic_models_routes.py` — pass the new fields through POST `/knowledge/concepts` and PATCH `/knowledge/concepts/{iri}`.
- `tests/integration/test_knowledge_routes.py` — new `TestKnowledgeConceptTypes` class with 10 cases covering create + read for every type combination and partial PATCH of property fields.

## Test plan

- [x] Backend integration tests for every concept type (`TestKnowledgeConceptTypes`, 10 cases) pass.
- [x] Live end-to-end Playwright walk-through of all 7 dialog options against a running stack:
  - Concept → badge `Concept`
  - Class → badge `Class`
  - Property + Object Property → badge `Property`, `DOMAIN: Person`, `RANGE: Organization`, `property_type=object`
  - Property + Datatype Property → badge `Property`, `RANGE: string` (from `xsd:string`), `property_type=datatype`
  - Property + Annotation Property → badge `Property`, no domain/range, `property_type=annotation`
  - Individual → `concept_type=individual`
  - Term → `concept_type=term` (recovered from `ontos:conceptType` annotation)
- [x] Cleanup: all seven verification concepts were deleted; API now returns 404 for each.


mvkonchits-db (approved):
LGTM.

Round-tripping `concept_type` / `property_type` / `domain` / `range` end-to-end fixes a real round-trip gap — Custom Properties were silently collapsing to `skos:Concept` on write. The `_resolve_concept_rdf_types` / `_detect_concept_type` symmetry is the right shape; centralising the mapping in one place keeps the create/update/promote/migrate paths honest. The `ontos:conceptType "term"` annotation as a sidecar for the SKOS-collision case is a pragmatic call.

- All 9 CI checks green (Backend, Frontend, TS, Build, Alembic single-head, Coverage, Title lint).
- 10-case `TestKnowledgeConceptTypes` covers create + read for every type combo plus partial PATCH of property fields.
- Live Playwright walkthrough of all 7 dialog options documented in the test plan, with cleanup verified.

One forward-looking note (non-blocking): #541 is stacked on this branch and will auto-retarget to `main` after merge — happy to take a look once CI re-runs on the retargeted base.

PR #544, opened 2026-06-19
fix(connectors): surface Snowflake connector credential fields

will-yuponce-db:
### Observed failure
In Settings -> Connectors -> Add Connection, selecting **Snowflake** shows NO credential/config fields and NO **Test Connection** button — only the generic Name / Description / Enabled / Default inputs. By contrast, selecting **BigQuery** renders a full configuration block (project, location, UC connection, secret scope/key, credentials path).

### Root cause
The connection form (`src/frontend/src/components/settings/connection-form-dialog.tsx`) is **not** data-driven from the connector metadata. It hardcodes a config block that is gated on `selectedType === 'bigquery'`, and `buildPayload` only collected `config` for `bigquery`. There was no Snowflake branch, so Snowflake fell through to the generic-only form.

This is purely a frontend gap, not a missing backend integration: the backend already advertises Snowflake config field hints (`account`, `user`, `warehouse`, `database`, `default_schema`, `role`) from `SnowflakeConnectorConfig` via `GET /api/connections/types` (`ConnectionsManager.list_connector_types`). The form simply ignored those hints.

(Note: the backend `SnowflakeConnector` is still a stub — `is_available = False` — so the underlying asset-discovery/Test integration remains pending. That is a separate, pre-existing backend limitation and out of scope for this UI defect. The **Test Connection** button itself is shown for any saved connection while editing; the original report's "no Test Connection button" was because the unsupported Snowflake form could not be meaningfully filled in/saved with config.)

### Fix
- Added a **Snowflake Configuration** block to the dialog mirroring the existing BigQuery pattern, exposing: Account Identifier, User, Warehouse, Role, Default Database, Default Schema.
- Extended the Zod form schema and the `form.reset` defaults to round-trip the Snowflake config fields when editing.
- Extracted the per-connector config-field mapping into a pure, exported helper — `CONNECTOR_CONFIG_FIELDS` + `buildConnectionConfig(connectorType, values)` — replacing the inline BigQuery-only loop in `buildPayload`. This keeps the logic data-driven and unit-testable, and prevents fields from one connector type leaking into another's persisted `config`.
- Added the `connectors.snowflake.*` i18n strings to `en/settings.json`.

### How tested
- New focused unit test `src/frontend/src/components/settings/connection-form-dialog.test.ts` (5 cases) covering `CONNECTOR_CONFIG_FIELDS` and `buildConnectionConfig` for Snowflake + BigQuery, including cross-connector field isolation and empty-field stripping. All pass via `vitest run`.
- `tsc --noEmit` type-check: clean.
- Full `src/components/settings/` vitest suite: passing (existing `role-form-dialog` test remains skipped, unchanged).
- JSON validity of `en/settings.json` confirmed.

The full dialog is not rendered in jsdom because its Radix UI Select hangs there (same documented limitation as `role-form-dialog.test.tsx`); the testable logic was therefore extracted into pure helpers.

### DIFF
--- a/src/frontend/src/components/settings/connection-form-dialog.test.ts
+++ b/src/frontend/src/components/settings/connection-form-dialog.test.ts
@@ -1,0 +1,94 @@
+/**
+ * Unit tests for connection-form-dialog config-field helpers.
+ *
+ * These cover the config-building logic that determines which credential /
+ * config fields are persisted per connector type. The full dialog itself
+ * relies on Radix UI components that hang in jsdom, so the testable logic is
+ * extracted into pure functions (see role-form-dialog.test.tsx for context).
+ *
+ * Regression: ONT-FEAT-029 — selecting "Snowflake" in Add Connection showed no
+ * credential/config fields because the Snowflake connector type was missing
+ * from the form's config-field map. These tests lock in that Snowflake exposes
+ * its credential/config fields just like BigQuery.
+ */
+import { describe, it, expect } from 'vitest';
+import {
+  CONNECTOR_CONFIG_FIELDS,
+  buildConnectionConfig,
+} from './connection-form-dialog';
+
+describe('CONNECTOR_CONFIG_FIELDS', () => {
+  it('exposes BigQuery config fields', () => {
+    expect(CONNECTOR_CONFIG_FIELDS.bigquery).toEqual([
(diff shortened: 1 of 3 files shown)


larsgeorge-db (approved):
lgtm

PR #545, opened 2026-06-19
fix(catalog-commander): allow applying a tag to a selected table

will-yuponce-db:
## CUJ
ONT-FEAT-007 — Catalog Commander: apply a tag to a selected table.

## Observed failure
In Catalog Commander, selecting a table (e.g. `samples.nyctaxi.trips`) gives a toolbar (View Data / Move / Delete / Rename / Ask Ontos / Info / Operations / Comments) and an Info panel showing only notes/links/documents. There was **no way to apply a tag** to the selected table.

## Root cause
The Info panel renders `EntityMetadataPanel`, which only covers notes, links, and documents. No control surfaced the tagging API. The backend already exposes a generic, free-form entity-tag assignment API:

- `GET  /api/entities/{entity_type}/{entity_id}/tags`
- `POST /api/entities/{entity_type}/{entity_id}/tags:set`
- `POST /api/entities/{entity_type}/{entity_id}/tags:add`
- `DELETE /api/entities/{entity_type}/{entity_id}/tags:remove`

`TagsManager.set_tags_for_entity` / `add_tag_to_entity` pass `entity_type` straight through to the repository with no whitelist, and `AssignedTagCreate` accepts a list of FQN strings or `{tag_fqn}` objects — exactly what the existing `TagSelector` produces. The Comments panel already uses `entity_type="catalog-object"` for these same items. So this was a **frontend surfacing gap**, not a missing feature — fully localizable, no backend change required.

## Fix
- New `EntityTagsPanel` (`src/frontend/src/components/metadata/entity-tags-panel.tsx`): loads assigned tags via the entity-tags GET endpoint, displays them with the existing `TagChip`, and lets the user edit/apply via the existing `TagSelector`, saving with `tags:set` (`[{ tag_fqn }]` payload). Editing is gated on the `tags` `READ_WRITE` permission, mirroring `TagSelector`'s own gating; without write access the panel is read-only.
- Wired the panel into the Catalog Commander Info panel for the selected object, using `entity_type="catalog-object"` (consistent with the Comments panel).
- Added `tags.*` i18n keys to `metadata.json` for all 7 locales (English copy as baseline for non-English).
- Icon buttons get `aria-label`s for accessibility/testability.

## Testing
- New unit test `entity-tags-panel.test.tsx` (vitest) — 3 cases, all pass:
  1. loads assigned tags from `GET /api/entities/catalog-object/{id}/tags`
  2. applies a tag via `POST .../tags:set` with `[{ tag_fqn }]` payload
  3. hides the edit control without `tags` write permission
- `tsc --noEmit` passes cleanly (exit 0, no errors).
- No backend change; the generic entity-tag endpoints already support arbitrary entity types.

### DIFF
--- a/src/frontend/src/components/metadata/entity-tags-panel.test.tsx
+++ b/src/frontend/src/components/metadata/entity-tags-panel.test.tsx
@@ -1,0 +1,137 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest'
+import { render, screen, waitFor } from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
+import { BrowserRouter } from 'react-router-dom'
+import { TooltipProvider } from '@/components/ui/tooltip'
+import { FeatureAccessLevel } from '@/types/feature-access-levels'
+
+// ---- Mocks ----------------------------------------------------------------
+
+const mockGet = vi.fn()
+const mockPost = vi.fn()
+
+vi.mock('@/hooks/use-api', () => ({
+  useApi: () => ({
+    get: mockGet,
+    post: mockPost,
+    put: vi.fn(),
+    patch: vi.fn(),
+    delete: vi.fn(),
+    loading: false,
+  }),
+}))
+
+const toastSpy = vi.fn()
+vi.mock('@/hooks/use-toast', () => ({
+  useToast: () => ({ toast: toastSpy }),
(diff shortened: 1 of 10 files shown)


larsgeorge-db (approved):
lgtm

PR #546, opened 2026-06-19
fix(contracts): hide New Contract action for Data Consumer

will-yuponce-db:
## CUJ
ONT-RBAC-007

## Observed failure
On the Data Contracts list page, the **New Contract** button was visible to a **Data Consumer** persona and the create dialog opened normally. The request only failed on save, where `POST /api/data-contracts` returned **403**. The unauthorized create action should never have been surfaced.

## Root cause
`src/frontend/src/views/data-contracts.tsx` rendered the **New Contract** and **Upload** create controls unconditionally — there was no frontend permission gate. The backend already enforces a 403, but the list view did not mirror the permission-gating pattern used elsewhere (e.g. the Data Products list), so read-only personas could open a dialog that was guaranteed to fail.

## Fix
Gate the create/upload controls behind the same permission check the Data Products list uses:

```ts
const { hasPermission, isLoading: permissionsLoading } = usePermissions();
const featureId = 'data-contracts';
const canWrite = !permissionsLoading && hasPermission(featureId, FeatureAccessLevel.READ_WRITE);
```

The **New Contract** and **Upload** controls now render only when `canWrite` is true. A Data Consumer (READ_ONLY) no longer sees them, so the action cannot be initiated. Backend 403 enforcement is unchanged (defense in depth).

## Testing
- Added `src/frontend/src/views/data-contracts.permissions.test.tsx` (vitest):
  - hides the New Contract / Upload controls for a READ_ONLY persona (Data Consumer)
  - shows them for a READ_WRITE persona
- `vitest run` for the new test plus existing `src/components/data-contracts/*` suites: **26 passed**.
- `tsc --noEmit` clean (no new type errors).

### DIFF
--- a/src/frontend/src/views/data-contracts.permissions.test.tsx
+++ b/src/frontend/src/views/data-contracts.permissions.test.tsx
@@ -1,0 +1,151 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import { render, screen, waitFor } from '@testing-library/react';
+import { MemoryRouter } from 'react-router-dom';
+import { FeatureAccessLevel } from '@/types/settings';
+
+// ---------------------------------------------------------------------------
+// CUJ ONT-RBAC-007 — the "New Contract" / "Upload" create actions must be
+// hidden for personas without contract-create (READ_WRITE) permission, e.g.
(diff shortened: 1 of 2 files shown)


larsgeorge-db (approved):
lgtm

PR #548, opened 2026-06-19
ci: add back-merge workflow to sync main into development

will-yuponce-db:
## Summary

Adds `.github/workflows/backmerge-main-to-development.yml` to keep the `development` branch in sync with `main` ("point main back down to development").

**Behavior**
- Triggers on every push to `main` (plus manual `workflow_dispatch`).
- Ensures an open PR exists merging **main → development**.
- **No-op** when `development` already contains `main` (`git merge-base --is-ancestor`).
- **Reuses** an existing open back-merge PR rather than creating duplicates (a branch-to-branch PR auto-tracks new commits on main).

**Why PR-based instead of a direct push/merge**
- Respects branch protection on `development`.
- If main and development have diverged, GitHub surfaces the conflicts on the PR for manual resolution instead of force-pushing over them.

**Permissions:** `pull-requests: write` only (uses the built-in `GITHUB_TOKEN`).

If you'd rather it auto-merge clean back-merges (no review), I can add `gh pr merge --auto` — left out by default so maintainers stay in control.

PR #552, opened 2026-06-22
ci: pin actions/checkout to full SHA in back-merge workflow

will-yuponce-db:
## Summary

The back-merge workflow added in #548 used `actions/checkout@v4`, which fails the org policy:

> Error: The action actions/checkout@v4 is not allowed in databrickslabs/ontos because all actions must be pinned to a full-length commit SHA.

Pins `actions/checkout` to the exact SHA the rest of this repo's workflows already use: `df4cb1c069e1874edd31b4311f1884172cec0e10 # v6`. No behavior change.

## Testing
- YAML validated; single-line change to the `uses:` pin.

### DIFF
--- a/.github/workflows/backmerge-main-to-development.yml
+++ b/.github/workflows/backmerge-main-to-development.yml
@@ -24,7 +24,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


mvkonchits-db (approved):
LGTM. Pure supply-chain hardening — pinning a third-party Action to a full SHA. Recommended practice. CI green, no runtime code touched.

PR #553, opened 2026-06-22
chore: hide alpha surfaces and mockup connectors from V1 UI

will-yuponce-db:
## Summary

Hides several not-ready surfaces from the default V1 UI so users only see GA/beta features and real connectors.

## Surfaces hidden

| Surface | Mechanism |
| --- | --- |
| Catalog Commander | maturity flipped `beta` -> `alpha` in `features.ts` (now hidden by default like the others) |
| Estate Manager | already `alpha`; now also removed from home Quick Actions |
| Entitlements | already `alpha`; now also removed from home Quick Actions |
| Security Features | already `alpha`; now also removed from home Quick Actions |
| Mockup connectors (Snowflake, PowerBI, Kafka) | filtered out of the Add Connection dropdown via `is_available` |

## Mechanism

**Maturity gating (4 navigation surfaces).** The sidebar already hides non-allowed maturities via `feature-visibility-store.ts` (default `showAlpha: false`), so `alpha` features are hidden from navigation. Two gaps closed:

1. `src/frontend/src/config/features.ts` — Catalog Commander changed from `beta` to `alpha` so it is hidden by default alongside Estate Manager / Entitlements / Security Features.
2. `src/frontend/src/components/home/quick-actions.tsx` — Quick Actions previously filtered **only** by permission, not maturity, so they still linked to alpha surfaces (Catalog Commander, Estate Manager, Entitlements) even when those were hidden from the sidebar. Quick Actions now also gate on `allowedMaturities` from `useFeatureVisibilityStore`, mirroring `navigation.tsx`, while keeping the existing permission filtering (both conditions combined).

**`is_available` filtering (connectors).** `src/backend/src/controller/connections_manager.py` `list_connector_types()` now skips connector types whose connector reports `is_available is False`. The Snowflake/PowerBI/Kafka stub connectors hard-code `is_available = False`, so they stop appearing in the Add Connection dropdown. The default connector type (Databricks) is always retained so the primary connection path is never hidden; real connectors like BigQuery still appear when available.

## Testing

- Frontend: `yarn type-check` (`tsc --noEmit`) — clean.
- Backend: added `src/backend/src/tests/unit/test_connector_types_availability.py` covering (a) stub connectors with `is_available=False` are excluded while available connectors are kept, and (b) the default connector is retained even when its `is_available` is `False`. Result: `2 passed`. (Run with a prebuilt dev env / coverage disabled; route integration tests `test_data_*_routes.py` fail on clean main due to a known audit-service fixture gap and are unrelated to this change.)

### DIFF
--- a/src/backend/src/controller/connections_manager.py
+++ b/src/backend/src/controller/connections_manager.py
@@ -199,8 +199,15 @@ class ConnectionsManager:
     # ------------------------------------------------------------------
 
     def list_connector_types(self) -> List[Dict[str, Any]]:
-        """Return metadata about all registered connector types."""
+        """Return metadata about registered connector types available for use.
+
+        Connectors that are not available (``is_available`` is False) are
+        filtered out so mockup/stub connectors (e.g. Snowflake, PowerBI, Kafka)
+        do not appear in the Add Connection dropdown. The default connector type
+        is always retained so the primary connection path is never hidden.
+        """
         registry = get_registry()
+        default_type = getattr(registry, "_default_connector_type", None)
         result = []
         for ctype in registry.list_registered():
             info: Dict[str, Any] = {
@@ -212,6 +219,10 @@ class ConnectionsManager:
             }
             try:
                 connector = registry.get_connector(ctype)
+                # Hide unavailable connectors (stubs hard-code is_available=False),
+                # but always keep the default connector type.
+                if not connector.is_available and ctype != default_type:
+                    continue
                 info["display_name"] = connector.display_name
                 info["description"] = connector.description
                 info["capabilities"] = {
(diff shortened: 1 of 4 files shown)


mvkonchits-db (approved):
LGTM.

- The \`is_available\` filter in \`list_connector_types\` is the right place to hide mockup stubs — single source of truth that feeds both the dropdown and any future programmatic callers.
- The default-connector-always-retained carve-out (\`ctype != default_type\`) is good defensive thinking. Prevents the user from ever ending up with an empty Add-Connection menu if the workspace client is transiently unattached.
- Wiring \`useFeatureVisibilityStore.allowedMaturities\` into \`quick-actions.tsx\` matches the sidebar gate exactly. Adding \`allowedMaturities\` to the \`useMemo\` deps is correct — without it the quick actions would stale on a maturity-pref change.
- Unit tests cover both filter branches.

CI green across the board. Approve.

PR #554, opened 2026-06-22
fix(data-products): actually gate consumer reads of unpublished products

will-yuponce-db:
## Summary

Follow-up to #535 ("gate direct reads of unpublished products (NEG-011)"), which was **merged and deployed but ineffective**: re-run on deployed main, a Data Consumer (`data-products=Read-only`, confirmed via `/api/user/permissions`) could still `GET /api/data-products/{draftId}` and receive `200` with the full draft (status `draft`). Expected `404`. Fixes **ONT-NEG-011**.

## Root cause: why #535 didn't take

The #535 gate (`_caller_can_read_product`) resolved the caller's project scope via `projects_manager.get_user_projects(...)`, then granted the read if the product appeared in `manager.list_products(...)` for that scope.

But `get_user_projects` has its own naive admin check:

```python
is_admin = any("admin" in group.lower() for group in user_groups) if user_groups else False
```

This treats **any** group whose name merely *contains* the substring `"admin"` — e.g. the ubiquitous workspace `account-admins` group, which is **not** the configured app-admin group (`APP_ADMIN_DEFAULT_GROUPS` defaults to `["admins"]`) — as a global admin and returns **every project**. A Data Consumer in such a group got `caller_project_ids = all projects`, which matched the draft's `project_id` inside `list_products`, so the gate returned `True` and leaked the draft.

Verified with a real-DB repro: with the old gate, a consumer in `account-admins` reads a draft in a project they do not belong to (`can_read_draft=True`); with the new gate they are denied (`False`).

## The fix

Stop trusting the broad listing scope. The gate now decides from **this product's own ownership facts**, using membership-based checks only:

- Published (`active`/`deprecated`) stays readable by everyone (catalog/marketplace contract).
- data-products feature admins (incl. in-app role override) see everything.
- Otherwise grant only **genuine ownership of this product**:
  - `draft_owner_id == caller` (case-insensitive, creator ownership),
  - `owner_team_id` in the caller's real team memberships (`teams_manager.get_teams_for_user`, membership-based, no substring-admin shortcut),
  - project membership via `projects_manager.is_user_project_member(...)`, which uses configured admin groups (`is_user_admin`) and real project-team membership rather than a substring match.

A denied read returns `404` (not `403`) so the draft's existence isn't disclosed.

## Testing

- Real-DB regression in `test_data_product_read_gate.py` proving a consumer in an `"admin"`-substring group (`account-admins`) is **denied** a draft in an unowned project; published-in-unowned-project stays readable; the draft owner is allowed.
- Updated the existing mock-based gate tests to the new ownership-fact logic.
- `test_data_product_read_gate.py`, `test_data_products_ownership_scope.py`, `test_data_products_repository.py`: **57 passed**.
- The `test_data_*_routes.py` integration suites and 7 `test_data_products_manager.py` cases fail identically on clean `main` (pre-existing audit-service fixture gap), unrelated to this change. Zero new failures introduced.

### DIFF
--- a/src/backend/src/routes/data_product_routes.py
+++ b/src/backend/src/routes/data_product_routes.py
@@ -104,36 +104,66 @@ def _caller_can_read_product(request, db, current_user, manager, product) -> boo
             if auth_manager.has_permission(eff, DATA_PRODUCTS_FEATURE_ID, FeatureAccessLevel.ADMIN):
                 return True
 
-        # Non-admin: resolve ownership scope and check the accessible set.
+        # Non-admin: grant only if the caller genuinely OWNS *this* product.
+        #
+        # ONT-NEG-011 follow-up (PR #535 was ineffective): the prior fix
+        # resolved the caller's scope via ``projects_manager.get_user_projects``
+        # and then checked membership of ``manager.list_products(...)``. But
+        # ``get_user_projects`` treats *any* group whose name merely contains
+        # the substring "admin" (e.g. the ubiquitous workspace ``admins``
+        # group) as a global admin and returns EVERY project. A Data Consumer
+        # in such a group therefore got ``caller_project_ids`` = all projects,
+        # which matched the draft's ``project_id`` in ``list_products`` and
+        # leaked the draft. We must NOT trust that broad, substring-admin
+        # scope. Instead resolve the three ownership facts against THIS product
+        # only, using membership-based checks (no substring-admin shortcut):
+        #   * draft_owner_id == caller (creator ownership)
+        #   * owner_team_id in caller's real team memberships
+        #   * project_id where the caller is a genuine project member
         from src.controller.teams_manager import teams_manager
         from src.controller.projects_manager import projects_manager
 
-        caller_team_ids: list = []
-        caller_project_ids: list = []
-        try:
-            user_teams = teams_manager.get_teams_for_user(db, caller_email, user_groups)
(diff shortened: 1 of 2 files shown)


mvkonchits-db (approved):
LGTM — and this is an important fix. PR #535 was structurally ineffective because \`projects_manager.get_user_projects\` over-grants on any group whose name *contains* the substring "admin" (e.g. the workspace-default \`account-admins\` group). A Data Consumer in such a group got every project's scope and matched the draft's \`project_id\` through \`list_products\`, exactly the NEG-011 leak.

The rewrite resolves the three ownership facts against THIS product only:
- \`draft_owner_id == caller\` (creator ownership, case-insensitive — good)
- \`owner_team_id\` in caller's real team memberships
- \`project_id\` where the caller is a genuine project member, via \`is_user_project_member\` which uses *configured* admin groups, not substring match

The \`manager.list_products.assert_not_called()\` line in \`test_draft_allowed_for_draft_owner\` is exactly the right kind of regression guard — it pins that the gate must never re-introduce the broad-listing shortcut.

The new \`TestProductReadGateSubstringAdminLeak\` class exercises the real gate against a real DB with a real \`DataProductsManager\`, which is the only way to actually pin this. The \`account-admins\` group choice in the test is the precise repro of NEG-011.

One forward-looking note (non-blocking): the substring-admin pattern in \`get_user_projects\` is still latent — anything else calling it will have the same bug. Worth filing a follow-up to tighten that helper or audit its remaining callers.

Approve.

PR #555, opened 2026-06-22
fix(ontology): emit skos:inScheme and support concept scheme membership

will-yuponce-db:
## Defect
ONT-ONTO-005 (regression follow-up). On deployed main, the collection Turtle export (`/api/knowledge/collections/{iri}/export?format=turtle`) had ZERO `skos:inScheme` triples and no `skos:ConceptScheme` declaration. Concepts were emitted as bare `skos:Concept` resources under the collection IRI namespace. Expected: a concept added to a collection/scheme is linked via `skos:inScheme` to a `skos:ConceptScheme`.

## Why the prior fix was ineffective
PR #526 (`fix(glossary): allow choosing collection in Create Concept dialog`) was purely frontend. It only changed the Create Concept dialog so the collection picker is no longer disabled when a default collection is passed. It did not touch the backend data model or the RDF export, so no `skos:ConceptScheme` was ever declared and no `skos:inScheme` triples were ever emitted. ONT-ONTO-005 therefore still failed after #526 merged.

## Root cause
`SemanticModelsManager.export_collection_as_turtle` / `export_collection_as_rdfxml` simply serialized the collection's named-graph context verbatim. That context contained the concept triples (`skos:prefLabel`, `ontos:status`, etc.) but never a `skos:ConceptScheme` declaration for the collection nor any `skos:inScheme` membership triples, because `create_concept` did not write them.

## Fix
- `create_concept` now persists a `skos:inScheme` triple (`concept -> collection IRI`) in both the DB (`rdf_triples` via the repo) and the in-memory graph, giving concepts proper scheme membership in the data model.
- A new `_build_collection_export_graph` helper backs both export formats. It copies the collection context, declares the collection IRI as a `skos:ConceptScheme` (carrying its `rdfs:label`), and adds `skos:inScheme` for every concept-typed subject in the context. Enriching at export time means collections created before this change also export correctly, not just newly created concepts.

The collection IRI doubles as the `skos:ConceptScheme` IRI (single-scheme membership), which delivers the export correctness ONT-ONTO-005 requires.

## Multi-scheme membership (needs-design, deferred)
True multi-scheme membership (a concept belonging to more than one scheme, with picker UI) is a larger design effort spanning the create/edit dialogs, the data model, and cross-collection promotion semantics. This PR delivers the single-scheme `skos:inScheme` linkage and export correctness now; multi-scheme UI should be tracked separately.

## Testing
- New unit tests `src/backend/src/tests/unit/test_collection_inscheme_export.py` (3 tests): assert `skos:inScheme` persistence on create, and that both Turtle and RDF/XML exports contain the `skos:ConceptScheme` declaration plus the `skos:inScheme` membership triple. All pass.
- Re-ran existing `test_semantic_models_manager_update.py`, `test_semantic_models_repository.py`, `test_rdf_serialization_display.py`: 22 passed, zero new failures.
- Note: `test_data_*_routes.py` integration tests fail on clean main due to a known audit-service fixture gap (pre-existing); validated via unit/manager tests.

References ONT-ONTO-005. Follows up #526.


mvkonchits-db (dismissed):
LGTM.

Nice belt-and-braces design:
- **Persistence path** writes \`skos:inScheme\` at concept-create time so new collections are self-describing.
- **Export-time enrichment** in \`_build_collection_export_graph\` back-fills \`skos:ConceptScheme\` + \`skos:inScheme\` for any subject typed as a SKOS / OWL / RDF concept-bearing class in the context. This means **pre-existing collections created before this change** still export correctly — no data migration needed.

The export builder uses a standalone \`Graph()\` and copies triples, so the in-memory \`_graph\` isn't polluted with the scheme declarations (they only live in the serialized output). Right call — keeps the export logic self-contained.

The member-types set (\`SKOS.Concept\`, \`OWL.Class\`, \`OWL.ObjectProperty\`, \`OWL.DatatypeProperty\`, \`OWL.AnnotationProperty\`, \`RDF.Property\`, \`OWL.NamedIndividual\`) matches the type set #540 just made first-class — these PRs compose cleanly.

Three tests (persistence, Turtle export, RDF/XML export) all parse the output back and assert on triples — that's the right shape for RDF tests.

Approve.

joshua-db:
Just a small comment on this:

> The collection IRI doubles as the skos:ConceptScheme IRI

Do we want to replace the "collection" terminology with "concept scheme"? They are essentially the same thing so it makes sense to pick one. Concept scheme is what the underlying resource is and it is also better understood (an RDF collection is actually something different) so I would suggest using that.

larsgeorge-db:
> Just a small comment on this:
> 
> > The collection IRI doubles as the skos:ConceptScheme IRI
> 
> Do we want to replace the "collection" terminology with "concept scheme"? They are essentially the same thing so it makes sense to pick one. Concept scheme is what the underlying resource is and it is also better understood (an RDF collection is actually something different) so I would suggest using that.

Ah, this is a good time to decide what we are doing here... so "Collections" are really more separate sources, or say separate RDF files. They can be downloaded separately, maintained separately, etc. On the other hand, Concept Schemes are a logical construct, a container or grouping mechanism for concepts. In essence, they are more a label/tag on elements, but not a physical boundary, as one RDF file can contain any number of schemes.

The general question to discuss here is: How do we map the physcial organisation versus the logical (structural or conceptual) one? If we want Ontos to become a place to manage both aspects (and I am seeing it as such), we need to decide how we store this and how we present it/make it accessible.

As for storage, we have the `rdf_triples` table holding the merged triples with extra info on the source, and RDFLib's Conjunctive Graph is using that to identify the partitions by source. We also have a separate set of tables that store the original RDF files and custom "schemes" created in the app, named `semantic_models` (I know, this is not a good choice by any means, but is a legacy naming issue). This seems reasonable. There are challenges to a) cache this efficiently for performance reasons, and b) keep the triples table in sync with the raw semantic models.

As for representation, we are currently either showing the files (in Settings - RDF Sources, and in the Concepts - Collections views), or the conjunctive graph (in Concepts - Concepts/Graph/...) with the ability to turn them on/off by source.

If we switch to a container solution, using SKOS ConceptScheme, then we are blurring the physical from the conceptual view. But schemes allow us to create better separation, one that is intentional by the model. As in, assume we map the custom Collections of the app 1:1 into `skos:ConceptScheme`, and then load an external OWL file with a set of such schemes in them, the UI would show those as distinct units, not the one by physical layout, which is lacking.

But, what are we doing with say an OWL/TTL file that does not define any schemes? Or one that has schemas, but also triples that are not assigned one as well? We are creating an implicit scheme (representing the source) and assigning that to all such triples? In fact, would we not just create a parent `skos:ConceptScheme` and assign ALL triples of the file to it, representing the source? Then we could, in the UI also filter not only by explicit schemes, but also by the implicit source ones.

aktungmak:
I like the idea of being able to keep track of where/when definitions were imported and then export certain parts. However, I don't think the file should be a hard boundary long term.

Using ConceptSchemes as a container (bearing in mind that Concepts/Properties can exist in more than one ConceptScheme) is a more common practice that using a file as the boundary. This is particularly important when larger ontologies are split into multiple files but are logically a single set of definitions.

ConceptSchemes can also be nested so this is more expressive that file boundaries.

> How do we map the physcial organisation versus the logical (structural or conceptual) one?

I don't think the physical organisation is as important really, given that we provide tools for exporting subsets of the definitions in Ontos. For example, the user can select a particular ConceptScheme or group of concepts/properties to export.

For keeping track of physical provenance, the existing solution should work well or we can just add documentation triples on the ConceptScheme. We could also make use of labelling predicates from PROV-O if it turned out that we need it later.

> As for representation, we are currently either showing the files (in Settings - RDF Sources, and in the Concepts - Collections views), or the conjunctive graph

I feel this is too complicated and I would recommend paring this back to just showing everything grouped by ConceptScheme.

> If we switch to a container solution, using SKOS ConceptScheme, then we are blurring the physical from the conceptual view. But schemes allow us to create better separation, one that is intentional by the model.

Agreed, that is why I feel it is better to push the physical separation into the background as an attribute that can be viewed but not a fundamental part of the UI.

> But, what are we doing with say an OWL/TTL file that does not define any schemes? Or one that has schemas, but also triples that are not assigned one as well?

At import, we should prompt the user to select a ConceptScheme to add concepts/properties to if they are missing an skos:inScheme definition. We could also provide the option to add ALL definitions in the file to a new/existing ConceptScheme.

larsgeorge-db:
Foundation for PRD #603 ([PRD]: SKOS ConceptScheme as primary glossary axis). #603 builds on this PR's `skos:inScheme` persistence, `ConceptScheme` export declaration, and collection↔scheme (1:1) identity to add Group-by-Scheme, a scheme/source segmented axis, read-only loaded schemes, and an admin assign-scheme action.

PR #556, opened 2026-06-22
fix(data-products): surface proposed products in the steward review queue

will-yuponce-db:
## ONT-CUJ-020 (Golden Path)

On a freshly-deployed ontos-dev, with #549 effective a data product can reach `proposed`/`under_review` (Request Certification works), but the submitted product never appears in the steward review/approvals queue (the queue shows "No results"). This PR fixes that surfacing bug (part a). Part (b) — a full product-review screen — is a missing feature and is noted as needs-design below.

## Root cause (part a)

`ApprovalsManager.get_approvals_queue()` (`src/backend/src/controller/approvals_manager.py`) built the products list with:

```python
q = db.query(DataProductDb).filter(DataProductDb.status == 'draft')
```

Since #549 (ONT-CUJ-019), `request-certify` advances a pre-review product (draft/sandbox) to `proposed` via `DataProductsManager.submit_for_review`. A product that has actually been submitted for review is therefore no longer `draft`, so it never appeared in the queue, while not-yet-submitted drafts were shown instead. The sibling contracts query already correctly filters `status.in_(['proposed', 'under_review'])`.

This is the real steward surface for proposed products/contracts: #529 ("surface proposed contracts to stewards") documented that proposed assets reach stewards through `GET /api/approvals/queue` (and the `ON_REQUEST_REVIEW` workflow trigger), not via `/data-asset-reviews`, which targets Unity Catalog assets and requires an explicit reviewer-assignment policy that does not exist yet.

## Fix

Mirror the contracts filter so products in `proposed`/`under_review` surface in the steward review queue.

## Tests

`src/backend/src/tests/unit/test_approvals_manager.py`:
- Updated `test_get_approvals_queue_with_products` and `test_get_approvals_queue_mixed` to assert the corrected lifecycle semantics (proposed/under_review surface; draft and active do not).
- Added `test_proposed_product_surfaces_after_request_certify` as a focused ONT-CUJ-020 regression.

Run: `pytest src/tests/unit/test_approvals_manager.py -q` -> 6 passed (against the worktree source via a provisioned dev env). Route integration tests (`test_data_*_routes.py`) are skipped here due to a known pre-existing audit-service fixture gap on clean main.

## Part (b) — needs-design (not in this PR)

There is no product-review screen: even in `under_review` the product detail shows deliverables only, with no per-policy compliance pass/fail and no quality-metric review panel. CUJ-020 expects the steward to open the product review and see mappings/metrics/compliance results. That is a genuinely large feature spanning a new backend review-aggregation surface (per-policy compliance results + quality metrics for a product) and new frontend, and it needs design (including a steward-assignment policy if reviews are ever to be routed to `/data-asset-reviews`). Filing/triaging that is left as a follow-up; this PR delivers only the localizable part (a).

### DIFF
--- a/src/backend/src/controller/approvals_manager.py
+++ b/src/backend/src/controller/approvals_manager.py
@@ -41,11 +41,19 @@ class ApprovalsManager:
         except Exception as e:
             logger.debug(f"Approvals queue: contracts query failed: {e}", exc_info=True)
         
-        # Products pending certification
+        # Products awaiting steward review (proposed or under_review)
         try:
             from src.db_models.data_products import DataProductDb
-            # ODPS v1.0.0: Query products in 'draft' status (awaiting approval to become 'active')
-            q = db.query(DataProductDb).filter(DataProductDb.status == 'draft')
+            # A product enters the steward review queue once it is submitted for
+            # review: request-certify transitions a pre-review product
+            # (draft/sandbox) to 'proposed' (see DataProductsManager.submit_for_review,
+            # ONT-CUJ-019). Mirror the contracts filter so proposed/under_review
+            # products surface to stewards. Filtering on 'draft' instead listed
+            # not-yet-submitted products and hid the ones actually awaiting
+            # review (ONT-CUJ-020).
+            q = db.query(DataProductDb).filter(
+                DataProductDb.status.in_(['proposed', 'under_review'])
+            )
             for p in q.all():
                 items['products'].append({
                     'id': p.id,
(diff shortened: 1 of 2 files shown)


mvkonchits-db (approved):
LGTM.

Sharp bisection of the regression: #549 made \`request-certify\` advance products from \`draft\` -> \`proposed\`, but the approvals queue filter still expected \`draft\`. Effect was inverted — not-yet-submitted drafts showed up, just-submitted products didn't. Mirroring the contracts filter (\`status.in_(['proposed', 'under_review'])\`) is the right fix and keeps the two queues in step.

Test changes are honest:
- \`test_get_approvals_queue_with_products\` now asserts the corrected lifecycle (proposed + under_review surface; draft + active don't) — that pins the contract, not just the line of code.
- \`test_proposed_product_surfaces_after_request_certify\` is a focused ONT-CUJ-020 regression — exactly the right shape (one product, post-submit state, must surface).
- \`test_get_approvals_queue_mixed\` updated to use \`proposed\` consistently.

CI green across all 9 checks. Part (b) — the product-review screen — is correctly scoped out as a separate design effort.

Approve.

PR #580, opened 2026-06-30
fix(workflows): surface on_behalf_of and consumer_principals in webhook payloads

### LINKED ISSUE #579: [Bug]: on_behalf_of wizard step result not threaded into AGR entity_data — ${context.on_behalf_of.*} stays literal
**Affects:** v1.0.0
**Component:** Approval wizard frontend (`approval-wizard-dialog.tsx`) + access grants backend (`access_grants_manager.py`)

---

### What happens

When a process workflow includes an `on_behalf_of` step, the user's selection (type + value) is submitted to the backend correctly, but it is **never aggregated into `collectedFieldsRef`** in the frontend. As a result, `onComplete` passes `wizard_data` to the access grants manager with `on_behalf_of` missing. The backend's `entity_data` therefore has no `on_behalf_of` key, and `${context.on_behalf_of.value}` (and `.type`, `.display`) render as literal strings in all webhook `body_template`s instead of resolving to the selected principal.

### Root cause

In `submitStep()` (non-preview path), only `user_action` step payloads are spread into `collectedFieldsRef.current`. The `on_behalf_of` step type is never handled:

```ts
// only this branch exists in v1.0.0:
if (currentStep.step_type === 'user_action' && ...) {
  collectedFieldsRef.current = { ...collectedFieldsRef.current, ...submissionPayload };
}
// on_behalf_of result is silently dropped here
```

### Fix

Add a parallel branch immediately after:

```ts
if (currentStep.step_type === 'on_behalf_of' && submissionPayload && typeof submissionPayload === 'object') {
  collectedFieldsRef.current = { ...collectedFieldsRef.current, on_behalf_of: submissionPayload };
}
```

---

### What happens

`${entity.consumer_principals}` in a webhook `body_template` renders as a full JSON array of objects (e.g. `[{"type":"group","value":"my-group"}]`) because the template engine JSON-serializes lists. There is no way to extract just the scalar value string using template syntax — `${entity.consumer_principals.value}` silently resolves to empty because `_walk_dict` does not traverse into lists.

Workflow authors who need just the group name (e.g. to pass a single AD group to an external provisioning runbook) have no usable template variable today.

### Fix

Expose a scalar shortcut `consumer_principals_value` alongside the existing array in `access_grants_manager.py`, set to the `value` field of the first principal:

```python
if "consumer_principals_value" not in entity_data and cp_list:
    first = cp_list[0]
    if isinstance(first, dict):
        entity_data["consumer_principals_value"] = first.get("value", "")
```

Then webhook templates can use:

```json
"data_product_group": "${entity.consumer_principals_value}"
```

which works identically to `${entity.data_product_name}` — plain string, no special handling.

---

Both fixes require no backend migrations. Bug 1 is frontend-only; Bug 2 is backend-only (one file).

mvkonchits-db:
## Summary

- **`workflow_executor.py`** — strip the `user:` prefix that PrincipalPicker serialises approver specs with (e.g. `user:foo@bar.com`). Without this, `_resolve_role_to_users` stored the prefixed string as the notification recipient so approval tasks were invisible in the admin UI.
- **`access_grants_manager.py`** — add `consumer_principals_value` scalar to `entity_data` (first CP group's `value` field). Allows webhook templates to use `${entity.consumer_principals_value}` to embed the UC access-group name directly instead of quoting a full JSON array.
- **`approval-wizard-dialog.tsx`** — thread the `on_behalf_of` step payload into `collectedFieldsRef` / `wizard_data` so the process workflow receives it in `entity_data` and `${context.on_behalf_of.value}` resolves correctly in webhook body templates.

Fixes #579.

## Test plan

End-to-end tested on `fevm-valcon-demo` workspace against a live httpbun.com echo endpoint:

- [x] Approval task appears in admin UI after submitting an access request (confirms `user:` prefix fix)
- [x] `on_behalf_of` renders as group name (not `null` / unresolved) — tested with both **"my group" dropdown** (`users`) and **free-text other** (`admins`) modes
- [x] `data_product_group` field in webhook body contains the UC group string `099_TREASURE-DSHA-WEU-OAI-MAYAI22-DEV_GDMP_10` (confirms `consumer_principals_value` fix)
- [x] All 6 workflow steps pass: notification → approval → approval response → webhook (HTTP 200) → notification → pass

This pull request was authored with Claude Code assistance.

### DIFF
--- a/src/backend/src/common/workflow_executor.py
+++ b/src/backend/src/common/workflow_executor.py
@@ -281,6 +281,10 @@ def _resolve_role_to_users(
                     return [(r.name, r.id)]
         return [(role_name, role.id if role else None)]
 
+    # Strip "user:" prefix emitted by the PrincipalPicker (e.g. "user:foo@bar.com").
+    if role_spec.startswith('user:'):
+        role_spec = role_spec[len('user:'):]
+
     if '@' in role_spec:
         # Pre-comma-split single email -- treat as a direct recipient.
         return [(role_spec, None)]
--- a/src/backend/src/controller/access_grants_manager.py
+++ b/src/backend/src/controller/access_grants_manager.py
@@ -241,6 +241,10 @@ class AccessGrantsManager:
                     # via the wizard.
                     if "consumer_principals" not in entity_data:
                         entity_data["consumer_principals"] = cp_list
+                    if "consumer_principals_value" not in entity_data and cp_list:
+                        first = cp_list[0]
+                        if isinstance(first, dict):
+                            entity_data["consumer_principals_value"] = first.get("value", "")
                     # Catalog + output_ports enrichment for webhook
                     # body_templates that need ${entity.catalogs} and
                     # ${entity.output_ports}. Helper preserves caller
--- a/src/frontend/src/components/workflows/approval-wizard-dialog.tsx
+++ b/src/frontend/src/components/workflows/approval-wizard-dialog.tsx
@@ -651,6 +651,11 @@ export default function ApprovalWizardDialog({
       if (currentStep.step_type === 'user_action' && submissionPayload && typeof submissionPayload === 'object') {
         collectedFieldsRef.current = { ...collectedFieldsRef.current, ...submissionPayload };
       }
+      // Thread on_behalf_of step result through so wizard_data carries it into
+      // entity_data for ${context.on_behalf_of.*} template resolution.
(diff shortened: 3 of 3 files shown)


mvkonchits-db:
Closing — reopening from a direct branch on databrickslabs/ontos to fix CI (fork PRs can't authenticate with JFrog).

Issue #570, opened 2026-06-24
[Bug]: "Copy from Team" → Assign opens an empty Assign Owner form (selected member not passed)


mvkonchits-db:
**Area:** Data Products → Owners panel (`OwnershipPanel`). Affects any entity that renders the shared ownership panel.
**Version:** v1.0.0 (`09b2b31`)

### Summary
When a data product has an **Owner Team** assigned, the Owners panel shows a **"Copy from Team"** button. It lists the team's members; clicking **Assign** on a member is meant to hand off to the "Assign Owner" flow pre-filled with that member. Instead, the Assign Owner dialog opens **empty**, so each member must be re-typed by hand — which defeats the purpose of the feature.

### Steps to reproduce
1. Open a Data Product that has an **Owner Team** assigned (with permission to assign owners).
2. In the **Owners** panel, click **Copy from Team**.
3. The team members are listed. Click **Assign** next to one.
4. The Copy-from-Team dialog closes and the **Assign Owner** dialog opens — but the User field is blank (no email/name pre-filled).

### Expected
The Assign Owner dialog opens pre-populated with the selected member's identifier/name (role still chosen by the user), matching the dialog's own copy: *"They will be assigned via the 'Assign Owner' flow where you can choose a role."*

### Root cause
`src/frontend/src/components/common/ownership-panel.tsx` — the per-member **Assign** handler only toggles dialog visibility and passes no member context:
```tsx
onClick={() => {
  setCopyFromTeamOpen(false);
  setAssignDialogOpen(true);
}}
```
`src/frontend/src/components/common/assign-owner-dialog.tsx` — `AssignOwnerDialog` exposes no prop to seed the owner (`open, onOpenChange, objectType, objectId, onSuccess`) and resets its internal email/name state to empty whenever it opens.

### Suggested fix
- Add optional `initialEmail` / `initialName` (and optionally `initialType`) props to `AssignOwnerDialog`; seed the form fields from them on open instead of always clearing.
- In `OwnershipPanel`, capture the clicked member in state (e.g. `pendingMember`) and pass `initialEmail={pendingMember?.member_identifier}` / `initialName={pendingMember?.member_name}` into `AssignOwnerDialog`.

### Related
PRD #431 (Consolidate Team and Ownership Model).

mvkonchits-db:
[...]

PR #588, opened 2026-07-02
fix(assets): scope Data Consumer asset visibility to DP-linked assets

### LINKED ISSUE #583: [Bug]: Data Consumer gets 403 on GET /assets/{id} even when asset is linked to an active Data Product
## Summary

A Data Consumer navigating to an asset detail page (`/assets/<id>`) receives a 403 even when:
- The asset is explicitly linked to a Data Product (via a Deliverable / OutputPort)
- That Data Product has `active` status

## Steps to reproduce

1. Create a Data Product with status `active`
2. Add an asset as a Deliverable (OutputPort → linked asset via `portHasTable` / `portHasView` etc.)
3. Switch to the **Data Consumer** role
4. Navigate directly to `/assets/<asset-id>` (e.g. by clicking the asset link on the DP detail page)

**Expected:** Asset detail page loads — consumer has access because the asset is attached to a published DP.  
**Actual:** 403 — "Asset is not linked to a Data Product accessible to this user."

## Root cause

`AssetsManager.resolve_accessible_asset_ids` (called from `GET /api/assets/{id}` consumer path) delegates to `DataProductsManager.list_products(is_admin=False)` with **no caller context** (no `caller_email`, no `caller_team_ids`, no `caller_project_ids`).

`DataProductsRepository.get_multi` has a fail-closed guard: when `is_admin=False` and all three scope inputs are `None`, it returns `[]` immediately rather than applying any status filter. This was intentional for the DP listing endpoint, but `resolve_accessible_asset_ids` relies on the old assumption that `list_products(is_admin=False)` returns at least the published (active/deprecated) DPs.

The result: `resolve_accessible_asset_ids` always returns an empty set for any non-admin caller → every asset is denied → 403.

## Affected file

`src/backend/src/controller/assets_manager.py` — `resolve_accessible_asset_ids` method.

## Proposed fix

The purpose of `resolve_accessible_asset_ids` is "which assets can a consumer see via the marketplace?" — i.e. assets of published DPs, regardless of caller ownership. The fix is to fetch all products with `is_admin=True` and filter locally to consumer-visible statuses:

```python
from src.common.version_visibility import is_visible_consumer

products = data_products_manager.list_products(
    skip=0, limit=10_000, is_admin=True,
)
products = [p for p in products if is_visible_consumer(p)]
```

This matches the catalogue/marketplace contract: published (active/deprecated) DP assets are readable by any caller with `data-products:READ_ONLY`.

mvkonchits-db:
## Summary

Closes #583, #570

### #583 — Data Consumer 403 on assets linked to active DPs

**Root cause:** `resolve_accessible_asset_ids` called `list_products(is_admin=False)` with no caller scope, which intentionally returns empty (fail-closed). This caused every asset lookup for a Data Consumer to return 403 — the accessible-asset set was always empty.

**Fix:** Switch to `list_products(is_admin=True)` then filter with `is_visible_consumer` so consumers see assets from any active/deprecated DP without requiring ownership.

Changed files:
- `src/backend/src/controller/assets_manager.py` — `resolve_accessible_asset_ids` now uses `is_admin=True` + `is_visible_consumer` filter
- `src/backend/src/routes/assets_routes.py` — routes now call manager methods (`resolve_accessible_asset_ids`, `is_asset_accessible`) instead of the deleted `data_product_asset_scope` module

### #570 — Assign Owner dialog pre-fills when triggered from Copy From Team

**Fix:** `OwnershipPanel` tracks the clicked member in `pendingMember` state and passes `initialEmail`/`initialName` to `AssignOwnerDialog`, which seeds the input fields via a `useEffect` when `open` changes.

Changed files:
- `src/frontend/src/components/common/assign-owner-dialog.tsx` — added `initialEmail`/`initialName` props + seeding `useEffect`
- `src/frontend/src/components/common/ownership-panel.tsx` — added `pendingMember` state, sets it on Assign click

## Test plan
- [ ] Data Consumer role → Assets → open any asset linked to an active Data Product → expect 200 not 403
- [ ] Data Product with Owner Team → Ownership panel → Copy from Team → Assign → dialog pre-fills with that member's email/name


mvkonchits-db:
Superseded — re-filing from upstream branch directly so CI secrets are available.

PR #592, opened 2026-07-03
fix(#583): scope Data Consumer asset visibility to DP-linked assets

mvkonchits-db:
## Summary
- Fixes "Asset is not linked to a Data Product accessible to this user" error for consumers browsing assets (#583)
- Root cause: `list_products(is_admin=False)` with no caller scope context returns empty (fail-closed design), causing all consumers to get an empty asset list
- Fix: list all products with `is_admin=True` then filter to consumer-visible statuses (`active`, `deprecated`) — consistent with what consumers can already see in the Data Products catalog

## Changes
- `assets_manager.py` (`resolve_accessible_asset_ids`): use `is_admin=True` + `is_visible_consumer` filter instead of `is_admin=False` which returns empty without caller scope

## Test plan
- [ ] As a Data Consumer, grant "Read only" access to an asset linked to a published Data Product
- [ ] Browse to the Assets page — verify the asset is visible and accessible (no 403)
- [ ] Verify assets NOT linked to any Data Product are still not visible to consumers
- [ ] Verify admin users still see all assets regardless

### DIFF
--- a/src/backend/src/controller/assets_manager.py
+++ b/src/backend/src/controller/assets_manager.py
@@ -5,6 +5,7 @@ from sqlalchemy.orm import Session
 from sqlalchemy.exc import IntegrityError
 
 from src.repositories.assets_repository import asset_type_repo, asset_repo, asset_relationship_repo
+from src.common.version_visibility import is_visible_consumer
 from src.models.assets import (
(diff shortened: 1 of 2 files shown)


larsgeorge-db (approved):
lgtm

PR #593, opened 2026-07-06
fix(genie): space creation from data products was broken end to end

nehmetohmedb:
## Summary

Clicking **Create Genie Space** on a data product failed silently: no space was created and the user received no feedback. Four independent defects combined:

1. **Wrong Genie API schema** (`genie_client.py`): the client sent `{display_name, tables[].full_name, instructions}` to `POST /api/2.0/genie/spaces`, but the API requires `title`/`description`/`warehouse_id` at the top level plus the space definition as a JSON string in `serialized_space` (`{version: 2, data_sources.tables[].identifier, instructions.text_instructions[].content}`), with `data_sources.tables` **sorted by identifier**. The request 400'd on the missing `warehouse_id` before anything else. `warehouse_id` now falls back to `settings.DATABRICKS_WAREHOUSE_ID`. Schema verified against the live API.

2. **Shadowed notification method** (`notifications_manager.py`): `create_notification` is defined twice in `NotificationsManager`; the later object-based definition `(notification, db)` shadows the async kwargs version, so the Genie flow's three notifications (started / ready / failed) all raised `TypeError: unexpected keyword argument 'user_id'` and were swallowed. The kwargs version is renamed `create_user_notification`; its four call sites (Genie x3, `projects_manager`) are migrated, and the Genie sites now pass `type=NotificationType.*` instead of the nonexistent `status=` kwarg. Same defect family as duplicate tool class definitions elsewhere.

3. **Schema-level identifiers sent as tables** (`genie_client.py`): `collect_datasets_from_products` fed `port.asset_identifier` straight to Genie, but that field is often schema-level (`catalog.schema`), which Genie rejects (`Invalid table identifier`). Table FQNs are now resolved from the port contract's schema objects (`physical_name`); `asset_identifier` is used only when it is a 3-part table name.

4. **Notifications rolled back in the background task** (`data_products_manager.py`): `notification_repo.create` only flushes, and the background task's sessions close without commit, so even the ready/failed notifications that survived (2) were silently rolled back. Explicit commits added.

## Verification

Tested end to end on a live Databricks Apps deployment: clicking the button now produces the "started" notification, creates the space (verified with a 3-port product resolving 6 table FQNs from its contracts), and delivers the "Genie Space Ready" notification with a working link to the space. The failure path was also exercised and delivers the error notification.


mvkonchits-db (approved):
lgtm

One concern worth flagging: serialized_space is an opaque internal JSON string, not a versioned API contract. The PR already pins "version": 2 which is good (the API will reject a stale version rather than silently misinterpret it), but this field has already drifted once (hence this PR). Two suggestions for hardening:

1. Add a comment on the payload construction calling out that serialized_space format is reverse-engineered, not documented — so the next person who touches this knows it's a fragility point, not stable contract.
2. Long-term: migrate to databricks-sdk's typed Genie client (w.genie.create_space(...)) instead of raw HTTP. Schema changes would then surface at SDK update time rather than via silent runtime failures

PR #596, opened 2026-07-07
fix(dfra): concept ownership + property links + RDF dedup + notifications idle

larsgeorge-db:
## Summary

Batch of fixes from customer report ("dfra"). All root causes reproduced locally against isolated servers (`:8100` / `:3100`) and verified end-to-end.

## Fixes

- **Owners on concept/term/property didn't render.** `business_owners /by-object`, `metadata /entities/{type}/{id}/…`, and `semantic-links /iri/…` used path params for IRI-valued ids, which the Databricks Apps proxy mangles by collapsing `%2F%2F` → `/` in path segments (see #536). Switched to `?iri=` / `?object_id=` / `?entity_id=` query params — inert to that transformation — matching the pattern #536 established. Frontend callers (`ownership-panel`, `entity-metadata-panel`, `linked-objects-panel`) updated.
- **Business concepts/properties disappeared from contract columns.** `data-contract-details.tsx` declared `propertyLinks` without a setter and never populated it. Added `fetchPropertySemanticLinks` (schema-level, one call per schema via new `/semantic-links/entity-prefix/{type}/{prefix}` endpoint) that replaces the slice per schema so removed assignments don't linger.
- **RDF triples duplicated unboundedly on re-import.** `_skolemize_bnode` embedded rdflib's random per-parse bnode id, so re-importing an ontology with blank nodes (OWL restrictions, SHACL shapes) never hit the `uq_rdf_triple` constraint. Now canonicalise the graph (RGDA1 `to_canonical_graph`) before persisting, so identical content produces identical rows and re-imports are true no-ops.
- **Threshold-gated bloat diagnostics on `RdfTriplesRepository`.** When the table exceeds `RDF_TRIPLES_DIAGNOSTIC_THRESHOLD` (default 30000), log a one-shot forensic snapshot: blank-node count, constraint-bypass check, source_type + context breakdown, per-(subject,predicate) churn. Needed because the customer's 64k-with-no-file case is not locally reproducible.
- **`_register_sources_as_collections` polluted the Collections list** by registering the dynamic recomputed `urn:app-entities` / `urn:semantic-links` contexts as KnowledgeCollections. Added them to the skip set.
- **"Assignment shown twice."** `list_for_iri` returned explicit DB links + inferred graph links with no dedup. Now deduped on `(entity_type, entity_id)`; explicit wins over the synthetic inferred link.
- **Data domains stuck loading** in Home Discovery when no domain named "Core" existed. Now falls back to first root domain, then first domain.
- **Notifications polling kept Lakebase compute warm 24/7** — no tab-visibility handling. Store now pauses the interval on `visibilitychange` and resumes (with an immediate fetch) on visible.
- **Dev ergonomics — parallel worktrees:** frontend/backend dev ports are now fully env-configurable so multiple git worktrees can run isolated servers without colliding on `3000`/`8000`. `vite.config.ts` reads `VITE_PORT` (server port) and `VITE_PROXY_TARGET` (proxy target); the `dev-backend` hatch script honors `BACKEND_PORT`/`BACKEND_HOST`. Documented in `CONTRIBUTING.md` ("Running Multiple Worktrees Side-by-Side") and mirrored into the LLM-agent instructions (`.cursor/rules/08-testing-and-deployment.mdc`, `CLAUDE.md`) so Cursor/Claude sessions in secondary worktrees don't restart or rebind the primary `:8000`/`:3000` servers.

- **Semantic caches were destroyed on every mutation and never repopulated (perf — the "still slow after redeploy" report).** Prod logs showed every dashboard request logging `Persistent cache not found for {stats,taxonomies,concepts}, computing live` against a 68,684-triple graph. Two defects: (1) `on_models_changed` ran `rebuild_graph_from_enabled` (which rebuilds both cache tiers) and then immediately `_invalidate_cache()`, deleting what it just built; (2) read paths recomputed live on a miss but never repopulated, so the miss recurred every request. Fix: drop the redundant post-rebuild invalidation, and add `_ensure_caches_warm()` so a cold read recomputes all tiers once from the (always-fresh) singleton graph via the existing atomic writer. Audited all 12 graph mutators for the invalidate-or-rebuild invariant so this does not reintroduce the historical "changes don't show" staleness (single-worker deploy, singleton read manager).

## Tests

- `src/backend/src/tests/unit/test_rdf_bnode_dedup.py` — re-import idempotency, stable bnode ids across parses, threshold-gated diagnostic firing + throttle + no-op-below-threshold + disabled-when-zero.
- `test_semantic_links_manager.py` — added `test_list_for_iri_dedups_explicit_and_inferred_same_entity`.
- `src/backend/src/tests/unit/test_semantic_cache_warm.py` — cold-read warms both cache tiers, warm reads don't recompute, `on_models_changed` leaves caches warm, invalidate→read rewarms.

## Test plan

- [ ] `hatch -e dev run pytest src/tests/unit/test_rdf_bnode_dedup.py src/tests/unit/test_semantic_links_manager.py` — new tests pass (2 pre-existing failures in `test_semantic_links_manager` are unrelated).
- [ ] Concept detail page with slash-bearing IRI (`/concepts/browser/urn%3Aglossary%3Atest-1%2Ftest-1.1-prop`) — assign an owner via Ownership panel, reload → owner persists.
- [ ] Contract detail (e.g. `00400001-0000-4000-8000-000000000001`, `customers` schema) — column concept badges render (customer_id → customerId/Unique Identifier, email → emailAddress/PII).
- [ ] Import a bnode-bearing TTL twice into an editable collection — second import returns `triples_imported: 0`; total row count stable across `POST /api/semantic-models/refresh-graph`.
- [ ] Home discovery `/marketplace` — domain graph auto-renders (no "Loading domains…" stuck state).
- [ ] Notifications polling: with a tab focused, `/api/notifications` hits every 60s; hide the tab → hits stop; return → immediate fetch then interval resumes.

## Non-code answers (for customer reply)

- **Business terms / LA via Assets:** same role as objects/features/concepts — use concepts for the pilot.
- **KPIs / metrics:** use concepts or add relations to a KPI object; dedicated KPI concept type noted as future ask.
- **"Domain" in concepts** is RDF/semantic terminology only; linking objects/properties to actual **data domains** is separate and being worked on.

## Follow-up (open)

The customer's **64k rdf_triples with no ontology file loaded** is *not* locally reproducible — `uq_rdf_triple` holds perfectly on the current schema (all 6 cols NOT NULL per `z8_fix_rdf_triple_nulls`, zero constraint bypass). Most likely cause: an older build predating `uq_rdf_triple` / `z8`'s NULL-coalesce, and/or the random-bnode vector now fixed. The diagnostics in this PR will identify the vector the next time it happens; ask the customer for `SELECT source_type, count(*) FROM rdf_triples GROUP BY 1 ORDER BY 2 DESC` + `alembic current` output.

### DIFF
--- a/.cursor/rules/08-testing-and-deployment.mdc
+++ b/.cursor/rules/08-testing-and-deployment.mdc
@@ -19,12 +19,11 @@ scope:
 
 - The user's primary worktree owns ports **8000/3000** — **NEVER restart those** and never bind your own servers to them.
 - When working in a *separate* worktree, start your **own** frontend + backend on **distinct ports** so you don't collide with (or accidentally proxy into) another branch's servers. Worktrees can share the same local `app_ontos` PostgreSQL DB; only the HTTP ports must differ.
-- All ports are env-var driven (no code edits):
-  - `BACKEND_PORT` (default `8000`), `BACKEND_HOST` (default `0.0.0.0`) — via the `dev-backend` hatch script.
+- Frontend port + proxy target are env-var driven; the backend port is passed straight to `uvicorn` (the `dev-backend` hatch script is hard-coded to 8000, so a second worktree runs uvicorn directly):
   - `VITE_PORT` (default `3000`) — Vite dev server.
   - `VITE_PROXY_TARGET` (default `http://localhost:8000`) — must point at *your* backend port.
 - Example second pair (backend 8100 / frontend 3100):
-  - Backend (from `src/`): `BACKEND_PORT=8100 yarn dev:backend`
+  - Backend (from `src/`): `hatch -e dev run uvicorn --app-dir backend src.app:app --reload --host=0.0.0.0 --port=8100`
   - Frontend (from `src/frontend/`): `VITE_PORT=3100 VITE_PROXY_TARGET=http://localhost:8100 yarn dev:frontend`
 - Full contributor-facing docs: `CONTRIBUTING.md` → "Running Multiple Worktrees Side-by-Side".
 
--- a/.cursor/rules/08-testing-and-deployment.mdc
+++ b/.cursor/rules/08-testing-and-deployment.mdc
@@ -15,6 +15,19 @@ scope:
 - Always run Python snippets via `hatch -e dev run ...` to pick up the project's environment.
 - Input validation: Pydantic on the backend, Zod / `react-hook-form` on the frontend.
 
+**Multiple worktrees (parallel LLM sessions)**
+
+- The user's primary worktree owns ports **8000/3000** — **NEVER restart those** and never bind your own servers to them.
+- When working in a *separate* worktree, start your **own** frontend + backend on **distinct ports** so you don't collide with (or accidentally proxy into) another branch's servers. Worktrees can share the same local `app_ontos` PostgreSQL DB; only the HTTP ports must differ.
+- All ports are env-var driven (no code edits):
+  - `BACKEND_PORT` (default `8000`), `BACKEND_HOST` (default `0.0.0.0`) — via the `dev-backend` hatch script.
+  - `VITE_PORT` (default `3000`) — Vite dev server.
+  - `VITE_PROXY_TARGET` (default `http://localhost:8000`) — must point at *your* backend port.
+- Example second pair (backend 8100 / frontend 3100):
+  - Backend (from `src/`): `BACKEND_PORT=8100 yarn dev:backend`
+  - Frontend (from `src/frontend/`): `VITE_PORT=3100 VITE_PROXY_TARGET=http://localhost:8100 yarn dev:frontend`
+- Full contributor-facing docs: `CONTRIBUTING.md` → "Running Multiple Worktrees Side-by-Side".
+
 **Testing**
 
 - Backend: `pytest` (tests under `src/backend/src/tests/` and top-level `e2e/`).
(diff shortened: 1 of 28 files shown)


mvkonchits-db:
Reviewed this while triaging a batch of field reports from a customer building their semantic layer on Ontos. Solid PR — the cache-warming fix in particular checks out: `rebuild_graph_from_enabled()` nulls all tiers up front and ends with `_build_persistent_caches_atomic()`, and `_ensure_caches_warm()` self-heals on a cold read, so dropping the post-rebuild `_invalidate_cache()` is safe (a failed rebuild degrades to recompute, not stale data). Test coverage is strong.

Two clarifications so nearby customer-reported bugs don't get closed as covered here — both are **out of scope for this PR** and stand as separate issues:

- **#662 (orphaned semantic links after contract re-import)** — this PR fixes stale property-link *display within a session* (the replace-per-schema logic in `fetchPropertySemanticLinks`) and RDF *triple* dedup, but does not purge orphaned `EntitySemanticLinks` rows when re-import mints a new contract UUID (`entity_id = {contract_id}#{schema}#{property}` changes, old rows remain). Distinct defect.
- **#665 (contract-property / contract-schema links not navigable)** — `navigateToEntity` in `linked-objects-panel.tsx` is untouched here (this PR only changes the fetch URL); the missing `data_contract_property` / `data_contract_schema` cases remain.

One small consistency nit: the new `/semantic-links/entity-prefix/{entity_type}/{entity_id_prefix:path}` route (and the `fetchPropertySemanticLinks` caller that sends `{contractId}#{schema}#` as a path segment) still uses a path param — the exact pattern the rest of this PR migrates away from to dodge the Apps proxy `%2F%2F` collapse. Safe today since contract UUIDs and schema names don't contain `//`, but a schema name with a slash would hit the same proxy bug. Might be worth a query param for consistency, or a comment noting why it's safe.

Minor: `max_interval = max(base_interval * 4, base_interval)` always equals `base_interval * 4` for positive base — the `max()` is dead (harmless).

mvkonchits-db (approved):

PR #597, opened 2026-07-07
feat(domains): multi-domain assignment for teams, contracts, products, and assets (#520)

### LINKED ISSUE #520: [PRD]: Multi-Domain Assignment for Teams, Data Contracts, Data Products, and Assets
# PRD: Multi-Domain Assignment for Teams, Data Contracts, Data Products, and Assets

## Concept Model

The relevant primitives:

- **Data Domain** — a hierarchical governance grouping (`data_domains` table; self-referential `parent_id`). Domains describe what the data is about (e.g., Sales, Finance, HR, Customer Master) and are used for browsing, discovery, ownership scoping, and tag propagation to Unity Catalog.
- **Domain-aware entities** — Teams, Data Contracts, Data Products, and Assets each carry a single, optional reference to one Data Domain today.
- **Primary domain** — the one domain a downstream system that only accepts a single value (ODCS export, Unity Catalog `data_domain` tag) should treat as authoritative.
- **Additional domains** — further domains the entity also belongs to. These participate fully in browsing, discovery, filtering, and search but are exposed to single-value integrations through extension fields.

These are distinct from the marketplace `publication_scope = "domain"` setting (which controls visibility, not assignment) and from the ODCS/ODPS YAML `domain` string (which is a single name field, exported from the primary).

## Problem Statement

Real assets, products, contracts, and teams routinely span more than one business domain, but Ontos forces every domain-aware entity into exactly one domain. This produces three concrete pains:

1. **Shared entities are misfiled or duplicated.** A `customers` table belongs to both Sales and Marketing; a "Customer 360" data product is consumed by Support, Sales, and Marketing. Today users either pick the "least wrong" domain (silently hiding the entity from the others) or duplicate the entity, which then drifts.
2. **Org structure cannot be modeled honestly.** A platform team that supports Finance and HR has to be filed under one or split into two teams. Cross-domain contracts (e.g., a master data contract used by several domains) cannot represent that fact in metadata.
3. **Discovery silently fails.** Users browsing the Sales domain do not see assets that are filed under Marketing but actually used in Sales workflows. Cross-domain reuse — one of the main reasons to run a data marketplace — becomes invisible.

The single-domain limit is also inconsistent across the four entity types: teams, contracts, and assets store `domain_id` (FK / soft-FK to `data_domains.id`); data products store a free-form `domain` string column that sometimes contains an ID and sometimes a name. Cleaning this up is a prerequisite for any cross-domain feature.

## Solution

Replace each entity's single domain reference with a polymorphic many-to-many relationship that designates exactly one primary domain alongside zero or more additional domains. Internally this is one junction table — `entity_domain_associations` — modeled on `entity_tag_associations`, with a `is_primary` boolean and a uniqueness constraint that enforces at most one primary per (entity_type, entity_id).

Single-value integrations (ODCS/ODPS export, Unity Catalog tag sync) consume the primary as their canonical value and emit additional domains through extension fields — `customProperties` for ODCS, multi-tag for UC. Filtering, search, browsing, and the marketplace are upgraded to "any of" semantics: an entity matches a domain filter when any of its assigned domains match (optionally including descendants, as marketplace already does today). Existing single-domain assignments migrate cleanly: every existing value becomes a primary record in the junction table.

The frontend introduces a single shared `DomainMultiSelector` component (popover + searchable badge list with a primary indicator) that replaces every existing single-domain `<Select>` and the legacy free-text input on the data product form.

### Multi-domain assignment

1. As a Data Steward, I want to assign a Data Product to multiple Data Domains, so that consumers in any of those domains can find it.
2. As a Data Steward, I want to assign a Data Contract to multiple Data Domains, so that contracts shared across business areas (e.g., master data) reflect their true scope.
3. As a Data Producer, I want to assign an Asset to multiple Data Domains, so that a shared table (e.g., `customers`) appears in every domain that uses it.
4. As an Admin, I want to assign a Team to multiple Data Domains, so that platform or cross-functional teams are visible to every domain they support.
5. As a Data Steward, I want to designate exactly one of an entity's domains as primary, so that downstream systems that need a single value have a deterministic answer.
6. As a Data Steward, I want the UI to clearly mark which domain is primary and let me change it without removing/re-adding domains, so that the primary choice is reversible and obvious.
7. As a Data Steward, I want to add or remove an additional domain without affecting the primary, so that I can iterate on domain breadth without re-confirming the canonical owner.
8. As a Data Steward, I want to remove all domains and leave an entity unassigned, so that draft or unclassified entities do not have to fake an affiliation.

### Discovery, search, and filtering

9. As a Data Consumer browsing a Data Domain detail page, I want to see all assets, contracts, products, and teams that include this domain (primary or additional), so that I see the full inventory of what the domain governs.
10. As a Data Consumer in the marketplace, I want to filter by one or more domains and have results show entities matching any of them, so that I can explore the union of multiple domains in one query.
11. As a Data Steward, I want the existing "include child domains" marketplace behavior to keep working with multi-domain entities, so that selecting a parent domain still surfaces entities assigned to its descendants.
12. As a Data Steward, I want list views (Teams, Contracts, Products, Assets) to display every assigned domain as a badge, with the primary visually distinguished, so that I can scan domain affiliation without opening a detail page.
13. As a Data Steward, I want to filter list views by domain and see entities where the domain is primary OR additional, so that the filter behaves like a tag filter.
14. As a Data Consumer searching the catalog, I want full-text and semantic search to match an entity by any of its domain names, so that domain-keyed searches return cross-domain entities.

### Forms and editing

15. As a Data Steward editing a Team, I want to pick multiple domains from a searchable popover and mark one as primary, so that team scope reflects reality.
16. As a Data Producer creating a Data Contract, I want the wizard to let me select multiple domains with one marked primary, so that the contract is correctly classified at creation time.
17. As a Data Producer creating a Data Product, I want a single, consistent domain selector that stores domain IDs (not free-text names), so that domain values are unambiguous and round-trip with import/export.
18. As a Data Producer viewing an Asset detail, I want to edit its domain assignments inline (not just see a read-only ID), so that I can correct or extend asset classification without leaving the page.
19. As a Data Steward, I want bulk imports of Assets to accept a `domain_ids` column (semicolon-separated, with the first treated as primary), so that I can onboard cross-domain inventories at scale.
20. As a Data Steward, I want bulk exports of Assets to emit `domain_ids` (primary first), so that round-tripping the export preserves all assignments.

### ODCS/ODPS exports

21. As a Data Producer exporting a Data Contract to ODCS, I want the standard `domain` field to receive the primary domain name, so that downstream tools that read ODCS see a valid single value.
22. As a Data Producer exporting a Data Contract to ODCS, I want any additional domains to be emitted under `customProperties` (e.g., `additionalDomains: [...]`), so that no information is lost in the export.
23. As a Data Producer exporting a Data Product to ODPS, I want the same primary-plus-additional convention applied to the ODPS export, so that the two formats behave consistently.
24. As a Data Producer importing an ODCS contract that uses the additional-domains custom property, I want those domains to be re-attached on import, so that round-tripping preserves multi-domain assignment.

### Unity Catalog tag sync

25. As a Data Steward, I want the UC tag sync to write the `data_domain` tag value as the primary domain, so that the existing convention is preserved for catalogs and consumers that read a single value.
26. As a Data Steward, I want additional domains to be written as separate UC tags (e.g., `data_domain_additional` per assignment), so that downstream UC tooling can still see the full set without breaking single-value consumers.

### Domain lifecycle

27. As an Admin deleting a Data Domain, I want the system to block the deletion if any entity has it as their primary domain, so that I do not silently leave entities without a canonical owner.
28. As an Admin deleting a Data Domain that is only ever an additional domain, I want the deletion to proceed and the corresponding association rows to be removed, so that domain cleanup does not require manual unassignment of every reference.
29. As an Admin, I want a domain detail page section that lists every entity (across types) assigned to the domain, distinguishing primary from additional, so that I can review impact before deletion.

### Migration and backward compatibility

30. As a System Operator, I want every existing single-domain assignment to be backfilled as a primary association during migration, so that no entity loses its current domain.
31. As a System Operator, I want the migration to resolve the data product `domain` column whether it currently stores a domain ID or a domain name, so that legacy rows are not lost.
32. As an Admin, I want the migration to log any unresolvable domain values (e.g., a stale name with no matching domain) so that I can manually reconcile them after upgrade.

### RBAC and visibility

33. As a User assigned to a team that spans multiple domains, I want my project visibility to consider all domains my teams belong to, so that cross-domain projects do not become invisible after the migration.

### Data model

- A new polymorphic junction table holds every domain assignment for every entity type, with a boolean primary flag and audit columns. The schema enforces at most one primary per (entity_type, entity_id) and uniqueness on (domain_id, entity_type, entity_id).
- The four scalar columns disappear: `teams.domain_id`, `data_contracts.domain_id`, `assets.domain_id`, `data_products.domain`. They are replaced entirely by the junction table.
- The migration backfills every non-null value as a primary association. For data products, the migration first attempts to interpret the legacy `domain` string as an ID, then falls back to a domain name lookup; unresolved values are logged and dropped.
- `data_domains` keeps its hierarchy. Multi-domain assignment is orthogonal to the parent/child tree — assigning a parent does not implicitly assign children.

### Backend modules

- **EntityDomainAssociationRepository** is the deep module. It provides: `set_domains_for_entity(entity_type, entity_id, domain_ids, primary_domain_id, assigned_by)` with replace-all semantics, idempotent upsert, batch reads (`get_domains_for_entities` for list endpoints to avoid N+1), and inverse lookups (`find_entity_ids_by_domain`, `find_entity_ids_by_domains`). Its interface mirrors `tags_repository.set_tags_for_entity`. It encapsulates the at-most-one-primary invariant and the deletion-blocking check.
- The four entity repositories (teams, contracts, products, assets) lose their domain column references and route domain reads/writes through the junction repository. Existing methods like `get_teams_by_domain`, `get_standalone_teams`, `get_by_domain`, `get_distinct_domains` are rewritten as junction-table queries with "any-of" semantics.
- The four entity managers gain a uniform `domain_ids: list[str]` plus `primary_domain_id: str | None` payload contract; managers validate the primary is in the set, resolve names→IDs where legacy code did so, and call the junction repository.
- A new **DomainExportAdapter** wraps the export-side concerns (ODCS `domain` field + `customProperties.additionalDomains`, ODPS equivalent, UC primary tag + additional tags). Both ODCS and UC code paths consume this adapter rather than reading the junction directly, keeping the export format conventions in one testable place.

### API contract

- Request/response payloads for the four entity types switch from `domain_id` (or `domain`) to `domain_ids: list[str]` and `primary_domain_id: str | null`. The primary is included in `domain_ids`. List endpoints accept `?domain_id=` (single, treated as any-of) and `?domain_ids=` (CSV, any-of) as filters.
- Existing teams routes that scope by domain (`/api/teams/domains/{domain_id}/teams`, `/api/teams/standalone`) keep their paths but adopt any-of semantics. "Standalone" means: zero domain assignments.
- The contract create payload no longer accepts the legacy `domain` (name) field; clients must pass `domain_ids` with resolved IDs.
- Domain delete endpoint surfaces a 409 with a structured body listing the entities for which it is primary, so the UI can offer reassignment guidance.

### Export and integration semantics

- **ODCS export**: standard `domain` string receives the primary domain's name; additional domains are emitted as a list under `customProperties` (key: `additionalDomains`, value: array of domain names). ODCS import recognizes the same custom property and reattaches the additional domains.
- **ODPS export**: same convention. Primary in the ODPS `domain` field; additional domains under `customProperties.additionalDomains`.
- **Unity Catalog tag sync**: primary written as the `data_domain` tag (current convention preserved). Each additional domain written as a separate tag with key `data_domain_additional`. The sync writes the primary tag first, then iterates the additionals.
- **Asset bulk CSV**: column renamed to `domain_ids`, semicolon-separated; the first ID in the list is the primary on import; on export the primary is emitted first, additional IDs follow.
- **Marketplace and term-mapping filters**: existing `RunTargetFilter.domain_ids: list[str]` already supports multi-value on the read side; adapters switch from filtering on the dropped scalar columns to joining through the junction table.

### Frontend modules

- **DomainMultiSelector** (new) is the deep frontend module: searchable popover, badge list with a primary indicator (e.g., a star icon or accent), add/remove and "set as primary" actions, controlled value `{ domain_ids, primary_domain_id }`. Modeled on the existing `TagSelector`. Used by every entity form.
- The data product create dialog and the legacy data product form both use the new selector. The legacy free-text `info.domain` input is removed.
- Asset detail gains a domain editor section using the same component.
- List views render badge groups in their domain columns. The `useDomains` hook gains a `getDomainNames(ids: string[])` helper.
- Marketplace and discovery filters become multi-select and pass `domain_ids` to the backend; the existing "include child domains" toggle behavior is preserved.

### RBAC and visibility

- The `projects_repository.get_projects_by_domain_relationship` join now derives the user's team-domain set from the junction table (union across all the user's teams' domains, primary and additional) and uses `IN` instead of equality. Project visibility expands consistently with multi-domain teams; no project-side schema change is needed for this PRD.

### Domain deletion

- Deletion is blocked when at least one entity has the domain as its primary. The 409 response includes the count and a small sample of affected entities per type so the UI can guide the user.
- When a domain is only ever an additional domain, deletion proceeds and association rows are removed.
- The existing `cascade="all, delete-orphan"` for `data_domains.children` is unchanged.

### Out-of-band cleanups bundled with this PRD

- `src/backend/src/file_models/data_products.py` references `entity.domain_id` against a model whose column is `domain` — broken today; replaced by reading from the junction table.
- Defensive code that matches data products by both ID and name (marketplace views) is removed once products store domain IDs only.

## Testing Decisions

Tests focus on external behavior — domain assignment outcomes, filter results, export shapes — not on implementation details like specific SQL or column names.

### Deep-module unit tests

- **EntityDomainAssociationRepository**: replace-all semantics (adding, removing, reordering domains), at-most-one-primary invariant (rejects two primaries, allows zero only when the set is empty), idempotent upserts, batch reads return correct primary marking, inverse lookups return the right entity IDs for any-of and primary-only queries, deletion-blocking returns the right blocker list. Prior art: `src/backend/src/tests/unit/test_data_domains_repository.py`.
- **DomainExportAdapter**: given a stub primary + additionals, produces the correct ODCS dict (primary in `domain`, additionals under `customProperties.additionalDomains`); UC tag sync emits the primary tag first and then one tag per additional; round-trip from ODCS dict back to (primary, additionals) is lossless. Prior art: `src/backend/src/tests/integration/test_odcs_export_validation.py`.

### Integration tests (per entity)

- For each of teams, contracts, products, assets: create with `domain_ids` + `primary_domain_id`, update to add/remove/swap primary, list with `?domain_ids=A,B` filter and assert any-of, list with `?domain_id=A` and assert any-of, GET detail returns the full set. Prior art: `test_data_contracts_db.py`, `test_teams_routes.py`, `test_data_product_routes.py`.

### Migration tests

- Seed each table with mixed legacy values, including: a contract with `domain_id`, a team with null `domain_id`, an asset with `domain_id`, a product whose `domain` is an ID, a product whose `domain` is a name, and a product whose `domain` is unresolvable. After the upgrade, assert: the right rows exist in `entity_domain_associations`, every backfilled row is `is_primary = true`, and unresolvable values are absent (and were logged).

### Frontend tests

- **DomainMultiSelector** unit tests: rendering with empty / single / multiple domains, add/remove flows, primary indicator, "set as primary" action, search filter, controlled value updates. Prior art: `src/frontend/src/components/ui/tag-selector.test.tsx` (if present) and `src/frontend/src/hooks/use-domains.test.ts`.
- Wizard/dialog updates: existing tests for `data-contract-wizard-dialog.test.tsx` and `data-product-create-dialog.test.tsx` updated to assert the new payload shape (`domain_ids`, `primary_domain_id`).

### What we deliberately do not test

- We do not snapshot SQL strings, column lists, or JSON shapes beyond the contract surface. Tests should keep passing if internal queries are rewritten as long as the API behavior and exports stay correct.

## Out of Scope

- **Per-domain RBAC and permissions.** Feature-level permissions (e.g., `data-domains`, `settings-data-domains`) are unchanged. Future work could add row-level domain-scoped permissions but is not part of this PRD.
- **Hierarchy-aware automatic assignment.** Assigning an entity to a parent domain does not implicitly assign it to descendants. Marketplace's "include child domains" filter remains a discovery-time feature.
- **Multiple primaries** or per-domain ordering beyond `assigned_at`. The model is one primary per entity, period.
- **Soft-deletion of domains** or any kind of "archive" status. Domain deletion is hard delete (blocked when in use as primary).
- **Renaming the entire `domain` concept** (for example to "subject area" or "business glossary domain"). Naming is unchanged.
- **Backfill heuristics for assets without a current `domain_id`** (e.g., infer from contract or product). Backfill is purely the literal column value.

## Further Notes

- The four scalar columns are dropped in the same migration that introduces the junction table. There is no transitional period where both shapes are accepted; clients update with the release.
- The data product `domain` column has long been a soft-typed string with mixed ID/name semantics. This PRD removes the ambiguity at the storage layer and at the API layer at the same time. UI code that previously matched products by both ID and name is simplified.
- The existing process workflow scoping (`scope_config: {type: "domain", ids: [...]}`) and term-mapping `RunTargetFilter.domain_ids` already use list-of-IDs shapes and require no contract changes; only the underlying joins move to the junction table.
- The plan file generated earlier (multi-domain_assignment_2f7410be.plan.md) reflects the prior "fully replace, no primary" decision and needs to be revised to add the `is_primary` column, the export adapter, and the deletion-block behavior described here.

dipankarkush-db:
## Summary

Implements **multi-domain assignment** (issue #520): replaces each entity's single domain reference — `teams.domain_id`, `data_contracts.domain_id`, `assets.domain_id`, and the free-text `data_products.domain` — with one polymorphic junction table, `entity_domain_associations`, modeled on `entity_tag_associations`, with an `is_primary` flag (exactly one primary per entity). The **primary** feeds single-value integrations (ODCS/ODPS `domain`, UC `data_domain` tag); **additional** domains ride extension fields (`customProperties.additionalDomains`, `data_domain_additional` UC tags) and participate in any-of search, filter, and discovery.

Closes #520.

## What changed

**Data model & migration**
- New `entity_domain_associations` junction (`is_primary`, uniqueness on `(domain_id, entity_type, entity_id)`, partial-unique primary per entity).
- Alembic `l1` migration backfills every existing single-domain value as a primary association (data-product `domain` resolved id-then-name; unresolvable values logged), then drops the four legacy columns.

**Backend**
- `EntityDomainAssociationRepository` (replace-all set, batch reads, any-of inverse lookups, deletion-block queries) and `DomainExportAdapter` (ODCS/ODPS + UC tag conventions).
- Teams, contracts, products, and assets wired to the junction (managers/repos/models/routes); MCP tools and the UC tag-sync workflow updated.
- Domain deletion blocked (409, structured impact body) when a domain is primary anywhere; `GET /data-domains/{id}/deletion-impact`.
- `?domain_ids=` (CSV, any-of) filter on all four list routes; search index matches any assigned domain name; project visibility unions all of a user's teams' domains.

**Frontend**
- Shared `DomainMultiSelector` (searchable, star marks primary) on team/contract/product/asset forms; `DomainBadgeList` on list rows and asset-detail.
- Domain-detail page lists assigned products/contracts/assets (primary vs additional); marketplace discovery matches any assigned domain; `useDomains.getDomainNames`.

## Verification
- Backend **1434** unit tests pass (1 pre-existing, unrelated network-hang); frontend **880** pass; `tsc` type-check clean.
- The migration was exercised on a real **Lakebase Postgres 16** instance — upgrade backfill (incl. uuid cast + product id/name resolution + unresolvable skip), downgrade restore, single alembic head, and the repositories end-to-end.
- Four `/code-review` (high-effort) passes; the final pass found no issues.

## Not in this PR (non-blocking)
- Per-entity integration tests + dialog-payload tests (blocked by the integration audit-fixture harness and Radix-in-jsdom; pure-logic tests added instead).
- Search-index full-reindex issues a per-entity domain query (perf-only, infrequent path).

### DIFF
--- a/.cursor/rules/10-entity-panel-matrix.mdc
+++ b/.cursor/rules/10-entity-panel-matrix.mdc
@@ -34,6 +34,7 @@ These tables use an `entity_type` + `entity_id` pattern to link to any entity.
 | `entity_relationships` | Typed relationships between any two entities |
 | `entity_subscriptions` | User subscriptions / notifications |
 | `entity_tag_associations` | Tag assignments |
+| `entity_domain_associations` | Data domain assignments — multi-domain with one `is_primary` per entity (`entity_type` ∈ {`team`, `data_contract`, `data_product`, `asset`}). Primary feeds single-value integrations (ODCS/ODPS `domain`, UC `data_domain` tag); additional domains ride extension fields and any-of discovery. See `docs/prds/prd-multi-domain-assignment.md`. |
 | `entity_semantic_links` | Links to ontology concepts/properties (`entity_type` ∈ {`data_domain`, `data_product`, `data_contract`, `data_contract_schema`, `data_contract_property`, `asset`, `uc_catalog`, `uc_schema`, `uc_table`, `uc_column`, `dataset` _(legacy)_}) |
 | `rich_text_metadata` | Rich-text notes |
 | `link_metadata` | External link bookmarks |
--- a/.github/workflows/test-coverage.yml
+++ b/.github/workflows/test-coverage.yml
@@ -2,9 +2,9 @@ name: Test Coverage
 
 on:
   push:
-    branches: [main, development]
+    branches: [main, develop, development]
   pull_request:
-    branches: [main, development]
+    branches: [main, develop, development]
   # Required by the merge queue: GitHub builds each batch on a temporary
   # gh-readonly-queue/<base>/... branch and fires a merge_group event. The
   # required status checks must run on that event or queued PRs hang forever
--- a/.github/workflows/test-coverage.yml
+++ b/.github/workflows/test-coverage.yml
@@ -2,9 +2,10 @@ name: Test Coverage
 
 on:
   push:
-    branches: [main, develop]
+    branches: [main, develop, development]
   pull_request:
-    branches: [main, develop]
+    branches: [main, develop, development]
+  merge_group:
 
 permissions:
   id-token: write
--- a/src/backend/alembic/versions/l1_add_entity_domain_associations.py
+++ b/src/backend/alembic/versions/l1_add_entity_domain_associations.py
@@ -1,0 +1,200 @@
+"""Add entity_domain_associations (multi-domain assignment) and drop legacy single-domain columns
+
+Introduces the polymorphic ``entity_domain_associations`` junction table (multi-domain
+assignment for teams, data contracts, data products and assets), backfills every existing
+single-domain value as a *primary* association, then drops the four legacy scalar columns
+(``teams.domain_id``, ``data_contracts.domain_id``, ``assets.domain_id``,
+``data_products.domain``).
+
+The data-product ``domain`` column is a free-text string that historically held either a
+domain ID or a domain name; the backfill resolves ID first, then name, and logs any values
+that cannot be resolved.
+
+Revision ID: l1_entity_domain_associations
+Revises: k1_merge_aa1_c1_g3
+Create Date: 2026-07-06
+"""
+import logging
+from typing import Sequence, Union
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects.postgresql import UUID as PG_UUID
+
+
+revision: str = "l1_entity_domain_associations"
+down_revision: Union[str, Sequence[str], None] = "k1_merge_aa1_c1_g3"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+logger = logging.getLogger("alembic.runtime.migration")
+
+TABLE = "entity_domain_associations"
+
+
+def _table_exists(conn, name: str) -> bool:
+    return conn.execute(
+        sa.text("SELECT 1 FROM information_schema.tables WHERE table_name = :n"),
+        {"n": name},
+    ).scalar() is not None
+
+
+def _column_exists(conn, table: str, column: str) -> bool:
+    return conn.execute(
+        sa.text(
+            "SELECT 1 FROM information_schema.columns "
+            "WHERE table_name = :t AND column_name = :c"
+        ),
+        {"t": table, "c": column},
+    ).scalar() is not None
+
+
+def upgrade() -> None:
+    conn = op.get_bind()
+
+    # 1. Create the junction table (idempotent guard for partial re-runs).
+    if not _table_exists(conn, TABLE):
+        op.create_table(
+            TABLE,
+            sa.Column("id", PG_UUID(as_uuid=True), primary_key=True),
+            sa.Column("domain_id", sa.String(), sa.ForeignKey("data_domains.id"), nullable=False),
+            sa.Column("entity_id", sa.String(), nullable=False),
+            sa.Column("entity_type", sa.String(), nullable=False),
+            sa.Column("is_primary", sa.Boolean(), nullable=False, server_default=sa.false()),
+            sa.Column("assigned_by", sa.String(), nullable=True),
+            sa.Column("assigned_at", sa.TIMESTAMP(timezone=True), server_default=sa.func.now(), nullable=False),
+        )
+        # Indexes mirror the ORM model (db_models/domain_associations.py).
+        op.create_index(f"ix_{TABLE}_domain_id", TABLE, ["domain_id"])
+        op.create_index(f"ix_{TABLE}_entity_id", TABLE, ["entity_id"])
+        op.create_index(f"ix_{TABLE}_entity_type", TABLE, ["entity_type"])
+        op.create_index(f"ix_{TABLE}_is_primary", TABLE, ["is_primary"])
+        op.create_index("ix_entity_domain_entity", TABLE, ["entity_type", "entity_id"])
+        op.create_unique_constraint(
+            "uq_entity_domain_assignment", TABLE, ["domain_id", "entity_type", "entity_id"]
+        )
+        # At most one primary domain per entity.
+        op.create_index(
+            "uq_entity_domain_primary",
+            TABLE,
+            ["entity_type", "entity_id"],
+            unique=True,
+            postgresql_where=sa.text("is_primary"),
+        )
+
+    # 2. Backfill existing single-domain values as PRIMARY associations.
+    #    Only rows whose domain resolves to a real data_domains.id are inserted
+    #    (satisfies the FK and drops stale references).
+    def _backfill_by_id(source_table: str, id_expr: str, domain_col: str, entity_type: str):
+        conn.execute(sa.text(
+            f"""
+            INSERT INTO {TABLE} (id, domain_id, entity_id, entity_type, is_primary, assigned_by, assigned_at)
+            SELECT gen_random_uuid(), s.{domain_col}, {id_expr}, :etype, TRUE, 'migration', now()
+            FROM {source_table} s
+            JOIN data_domains d ON d.id = s.{domain_col}
+            WHERE s.{domain_col} IS NOT NULL
+            ON CONFLICT ON CONSTRAINT uq_entity_domain_assignment DO NOTHING
+            """
+        ), {"etype": entity_type})
+
+    _backfill_by_id("teams", "s.id", "domain_id", "team")
+    _backfill_by_id("data_contracts", "s.id", "domain_id", "data_contract")
+    _backfill_by_id("assets", "CAST(s.id AS VARCHAR)", "domain_id", "asset")
+
+    # data_products.domain is free-text (ID or name). Resolve ID first, then name.
+    # (a) domain value IS a domain id
+    conn.execute(sa.text(
+        f"""
+        INSERT INTO {TABLE} (id, domain_id, entity_id, entity_type, is_primary, assigned_by, assigned_at)
+        SELECT gen_random_uuid(), d.id, p.id, 'data_product', TRUE, 'migration', now()
+        FROM data_products p
+        JOIN data_domains d ON d.id = p.domain
+        WHERE p.domain IS NOT NULL
+        ON CONFLICT ON CONSTRAINT uq_entity_domain_assignment DO NOTHING
+        """
+    ))
+    # (b) domain value is a domain NAME (and was not already resolved as an id)
+    conn.execute(sa.text(
+        f"""
+        INSERT INTO {TABLE} (id, domain_id, entity_id, entity_type, is_primary, assigned_by, assigned_at)
+        SELECT gen_random_uuid(), d.id, p.id, 'data_product', TRUE, 'migration', now()
+        FROM data_products p
+        JOIN data_domains d ON d.name = p.domain
+        WHERE p.domain IS NOT NULL
+          AND NOT EXISTS (SELECT 1 FROM data_domains d2 WHERE d2.id = p.domain)
+        ON CONFLICT ON CONSTRAINT uq_entity_domain_assignment DO NOTHING
+        """
+    ))
+    # (c) log unresolvable data-product domain values for manual reconciliation.
+    unresolved = conn.execute(sa.text(
+        """
+        SELECT p.id, p.domain
+        FROM data_products p
+        WHERE p.domain IS NOT NULL
+          AND NOT EXISTS (SELECT 1 FROM data_domains d WHERE d.id = p.domain)
+          AND NOT EXISTS (SELECT 1 FROM data_domains d WHERE d.name = p.domain)
+        """
+    )).fetchall()
+    for row in unresolved:
+        logger.warning(
+            "Multi-domain migration: could not resolve data_product %s domain value %r; dropped.",
+            row[0], row[1],
+        )
+
+    # 3. Drop the four legacy scalar columns (Postgres cascades FK/index cleanup).
+    if _column_exists(conn, "teams", "domain_id"):
+        op.drop_column("teams", "domain_id")
+    if _column_exists(conn, "data_contracts", "domain_id"):
+        op.drop_column("data_contracts", "domain_id")
+    if _column_exists(conn, "assets", "domain_id"):
+        op.drop_column("assets", "domain_id")
+    if _column_exists(conn, "data_products", "domain"):
+        op.drop_column("data_products", "domain")
+
+
+def downgrade() -> None:
+    conn = op.get_bind()
+
+    # 1. Re-add the legacy columns (nullable).
+    if not _column_exists(conn, "teams", "domain_id"):
+        op.add_column("teams", sa.Column("domain_id", sa.String(), sa.ForeignKey("data_domains.id"), nullable=True))
+    if not _column_exists(conn, "data_contracts", "domain_id"):
+        op.add_column("data_contracts", sa.Column("domain_id", sa.String(), sa.ForeignKey("data_domains.id"), nullable=True))
+    if not _column_exists(conn, "assets", "domain_id"):
+        op.add_column("assets", sa.Column("domain_id", sa.String(), nullable=True))
+    if not _column_exists(conn, "data_products", "domain"):
+        op.add_column("data_products", sa.Column("domain", sa.String(), nullable=True))
+
+    # 2. Restore the primary domain into each legacy column.
+    conn.execute(sa.text(
+        f"""
+        UPDATE teams t SET domain_id = a.domain_id
+        FROM {TABLE} a
+        WHERE a.entity_type = 'team' AND a.entity_id = t.id AND a.is_primary = TRUE
+        """
+    ))
+    conn.execute(sa.text(
+        f"""
+        UPDATE data_contracts c SET domain_id = a.domain_id
+        FROM {TABLE} a
+        WHERE a.entity_type = 'data_contract' AND a.entity_id = c.id AND a.is_primary = TRUE
+        """
+    ))
+    conn.execute(sa.text(
+        f"""
+        UPDATE assets s SET domain_id = a.domain_id
+        FROM {TABLE} a
+        WHERE a.entity_type = 'asset' AND a.entity_id = CAST(s.id AS VARCHAR) AND a.is_primary = TRUE
+        """
+    ))
+    conn.execute(sa.text(
+        f"""
+        UPDATE data_products p SET domain = a.domain_id
+        FROM {TABLE} a
+        WHERE a.entity_type = 'data_product' AND a.entity_id = p.id AND a.is_primary = TRUE
+        """
+    ))
+
+    # 3. Drop the junction table.
+    if _table_exists(conn, TABLE):
+        op.drop_table(TABLE)
--- a/src/backend/src/common/compliance_entities.py
+++ b/src/backend/src/common/compliance_entities.py
@@ -216,14 +216,22 @@ class AppEntityLoader(EntityLoader):
             # Data Products
             if 'data_product' in entity_types:
                 from src.repositories.data_products_repository import data_product_repo
+                from src.repositories.entity_domain_association_repository import entity_domain_repo
                 products = data_product_repo.list(self.db)
+                # Domain moved to the entity_domain_associations junction; batch-load the
+                # primary domain name per product (the legacy `domain` column was dropped).
+                domains_map = entity_domain_repo.get_domains_for_entities(
+                    self.db, entity_type="data_product", entity_ids=[str(p.id) for p in products]
+                ) if products else {}
                 for product in products:
+                    assigned = domains_map.get(str(product.id), [])
+                    primary_domain = next((a.domain_name for a in assigned if a.is_primary), None)
                     yield {
                         'type': 'data_product',
                         'id': product.id,
                         'name': product.name,
                         'description': product.description,
-                        'domain': product.domain,
+                        'domain': primary_domain,
                         'status': product.status,
                         'version': product.version,
                         'owner': product.owner,
--- a/src/backend/src/common/genie_client.py
+++ b/src/backend/src/common/genie_client.py
@@ -227,7 +227,8 @@ def collect_rich_text_metadata(product_ids: List[str], db: Session) -> Dict[str,
 def format_metadata_for_genie(
     metadata_map: Dict[str, List],
     products: List,
-    max_length: int = 5000
+    max_length: int = 5000,
+    product_domains: Optional[Dict[str, str]] = None,
 ) -> str:
     """
     Format collected metadata as markdown instructions for Genie.
@@ -236,6 +237,9 @@ def format_metadata_for_genie(
         metadata_map: Dict of entity_id -> metadata entries
         products: List of DataProductDb objects
         max_length: Maximum character length (default 5000)
+        product_domains: Optional {product_id: primary domain name}. Domain moved to the
+            entity_domain_associations junction (the legacy DataProductDb.domain column was
+            dropped), so the caller resolves it and passes it in here.
 
     Returns:
         Formatted markdown string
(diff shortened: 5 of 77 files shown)


dipankarkush-db:
## Isaac Review — Multi-domain assignment (#520)

**Verdict: Approve**, with two non-blocking follow-ups.

### Overview
Replaces the single-domain reference on teams / data contracts / data products / assets with one polymorphic junction table (`entity_domain_associations`, `is_primary`, at-most-one-primary). Primary feeds single-value integrations (ODCS/ODPS `domain`, UC `data_domain` tag); additional domains use extension fields + any-of discovery. Alembic migration backfills existing values as primary and drops the four legacy columns; full frontend (`DomainMultiSelector`, badges, wired forms, domain-detail inventory).

### Strengths
- Clean fit to the controller/repository/manager pattern; logic centralized in two deep modules (`EntityDomainAssociationRepository`, `DomainExportAdapter`) mirroring the `entity_tag_associations` blueprint.
- Correctness hardened over multiple passes: request-session threading for create/update/clone domain read-back; UUID-safe empty filter (`in_([])`); `AssetSummary` carries domain fields (fixed an asset-edit data-loss); ODPS form-dialog keeps domains out of the AJV `additionalProperties:false` payload.
- Migration backfill (product id-then-name, unresolvable logged, uuid cast) + downgrade **exercised on real Lakebase Postgres 16**.
- Conventions: single alembic head; new table documented in the entity-panel matrix; `%`-style logging.

### Test coverage
Added: repo (16), export adapter (9), delete-block (7), `DomainMultiSelector` logic (11), `getDomainNames`, Postgres-gated migration test. Backend **1434** pass, frontend **880** pass, type-check clean.

### Non-blocking follow-ups
1. **Search-reindex N+1** — `get_search_index_items` (assets/contracts/products) queries domains per-entity during a full reindex; perf-only, infrequent path; batch helper exists.
2. **Integration/dialog tests** — per-entity route + dialog-payload tests not included (blocked by the audit-fixture harness + Radix-in-jsdom).

### Risk
Breaking by design (per the issue): drops 4 columns; API switches to `domain_ids`/`primary_domain_id` with no transition window — migrate external consumers of the old fields.

_Reviewed by Isaac._

dipankarkush-db:
## Isaac Review (refresh @ eb822be8) — Multi-domain assignment (#520)

_Updated review covering the latest commits (CI-workflow fix + search-index batching); supersedes my earlier comment._

**Verdict: Approve.** No blocking issues; the two follow-ups from the prior review are now addressed.

### Overview
Replaces the single-domain reference on teams / contracts / products / assets with one polymorphic junction (`entity_domain_associations`, `is_primary`, at-most-one-primary). Primary feeds single-value integrations (ODCS/ODPS `domain`, UC `data_domain` tag); additional domains use extension fields + any-of discovery. Alembic migration backfills existing values as primary and drops the four legacy columns; full frontend.

### Correctness
- Hardened across five review passes: request-session threading for create/update/clone domain read-back; UUID-safe empty filter (`in_([])`); `AssetSummary` carries domain fields (fixed an asset-edit data-loss); ODPS form-dialog keeps domains out of the AJV `additionalProperties:false` payload; `?domain_ids=` any-of on all four list routes.
- Migration (backfill + downgrade) **and** the new `get_all_assets`/search-index batch paths verified on **real Lakebase Postgres 16**, not just SQLite.

### Performance
- Prior review's **search-reindex N+1 is now fixed** (commit `eb822be8`): `get_search_index_items` for assets/contracts/products batch-loads domains via `get_domains_for_entities` and passes them into `_build_search_index_item` (single-entity fallback retained for incremental updates).

### Conventions / CI
- Single alembic head; new table documented in the entity-panel matrix; `%`-style logging.
- Also fixes a repo CI bug (commit `9447e1a0`): `test-coverage.yml` only triggered on `main`/`develop`, so its required checks never ran for `development` PRs. Now includes `development` + `merge_group`. All required checks are green on this PR.

### Test coverage
Added: repo (16), export adapter (9), delete-block (7), DomainMultiSelector logic (11), getDomainNames, Postgres-gated migration test. Backend **1434** pass, frontend **880** pass, type-check clean.

**Remaining gap (non-blocking, pre-existing infra):** per-entity route integration + dialog-payload tests aren't included — the integration suite needs an audit-fixture harness (503s standalone) and Radix dialogs can't render in jsdom. Best as a follow-up.

### Risk
Breaking by design (per the issue): drops 4 columns; API switches to `domain_ids`/`primary_domain_id` with no transition window — migrate external consumers of the old fields.

_Reviewed by Isaac._

mvkonchits-db:
**Merge conflicts + a review pass while I was in here.** Rebased against `development` at `acdce21d`.

**Conflicts (3 files):**

`.github/workflows/test-coverage.yml` — trivial. `development` added merge-queue support (`merge_group:` + the `gh-readonly-queue` comment), your side added `develop` to the branch list. Take `development`'s version, add `develop` back if you want it. The merge-queue plumbing is what gates PRs into `main` now, so don't lose it.

`src/backend/src/controller/assets_manager.py` — trivial. Both sides added a different import; keep both:
```python
from src.repositories.entity_domain_association_repository import entity_domain_repo
from src.common.version_visibility import is_visible_consumer
```

`src/backend/src/controller/data_contracts_manager.py` — needs code, not a side-pick. Two regions:
- `_query_contracts(...)` call (~line 6221): `development` passes `status=`, your branch passes `domain_ids=`. Merged signature takes both, so keep both.
- `create_contract` (~line 2284): `development` added a duplicate-name guard (ONT-NEG-002, via back-merge #615) that runs `db.query(DataContractDb).filter(DataContractDb.domain_id == domain_id)`. Your PR drops that column, so "accept both" compiles but throws `AttributeError` on the first contract create. The guard has to be re-expressed against the junction table: resolve the primary via `entity_domain_repo`, then check name collision within that domain. Your `_resolve_domain_assignment` already hands you the primary.

CI on the PR is from Jul 8, before the 19 commits now on `development`. Please rerun backend tests after the rebase, especially contract create and the ONT-NEG-002 case.

**Review notes (non-blocking, spotted while resolving):**

The core is in good shape. Junction table, one-primary constraint, backfill, deletion-block, frontend selector all check out. A few gaps against the PRD:
1. ODPS product re-import is lossy. Export writes camelCase `domainIds`/`primaryDomainId` + `customProperties.additionalDomains`, but `_resolve_product_domain_assignment` reads only snake_case `domain_ids` and the single `domain` name, so additional domains drop on round-trip. Contracts round-trip fine through `domain_export_adapter.parse_odcs`. Routing product import through the same adapter would fix it.
2. UC additional-domain tags: PRD asks for one `data_domain_additional` tag per extra domain; `uc_tag_sync.py` comma-joins them into a single value and bypasses the adapter's `uc_tags()`, which already emits them separately.
3. N+1 in `list_products`: it calls `get_domains_for_entity` per row (`data_products_manager.py:433` -> `_attach_domains`), while every sibling list path batches via `get_domains_for_entities`. Perf-only.
4. Marketplace filter control is single-select; matching is any-of but the UI can't pick multiple domains (story 10).

None of these block merge once the conflicts are resolved.

dipankarkush-db:
Thanks for the rebase and the review pass, @mvkonchits-db — really helpful. Rebased onto `development` (now `acdce21d`); PR shows mergeable again. Latest is `5137c6d9`.

**Conflicts**
- `test-coverage.yml` — took `development`'s merge-queue version (`merge_group:` + comment), re-added `develop` to the branch lists.
- `assets_manager.py` — kept both imports.
- `data_contracts_manager.py` — `_query_contracts` call keeps both `status=` and `domain_ids=`. The **ONT-NEG-002 guard** is re-expressed against the junction: for each case-insensitive name match I resolve its primary via `entity_domain_repo.get_primary_domain_id` and compare to the new contract's resolved primary (unassigned → both `None` → still blocks in "global scope", matching the original). `test_contract_duplicate_name.py` passes.

**Review notes**
1. **ODPS product re-import** — fixed. Non-snake_case payloads now route through `domain_export_adapter.parse_odcs`, so `domainIds`/`primaryDomainId` + `customProperties.additionalDomains` round-trip. Kept a legacy `domain`-as-ID short-circuit (parse_odcs resolves `domain` only by name). Added 5 unit tests.
2. **UC additional-domain tags** — fixed, with one wrinkle worth calling out. A UC securable holds one value per tag key, so additional domains can't share a single `data_domain_additional` key. I emit them as **numbered keys** instead — `data_domain_1`, `data_domain_2`, … primary first — in both `uc_tag_sync` and `DomainExportAdapter.uc_tags`. One known limitation: `aggregate_parent_desired` (schema/catalog rollup) unions by key, so numbered additionals from different datasets can group by index. I left it as-is because that rollup is already documented-approximate and the *primary* tag is comma-collapsed there the same way; table-level tags are correct. Happy to revisit if you'd prefer a different parent-level scheme.
3. **`list_products` N+1** — fixed. Batches via `get_domains_for_entities` and threads through `_load_product_with_tags`/`_attach_domains`, like the sibling paths.
4. **Marketplace single-select** — fixed. `selectedDomainId` → `selectedDomainIds[]` with any-of union (children behavior preserved); graph focus tracked separately so browsing doesn't mutate the filter.

Backend unit suite green (1304 passed). Ran `/code-review` on the delta, which caught the legacy-`domain`-as-ID drop above (now fixed + tested).

dipankarkush-db:
@mvkonchits-db all your review notes are addressed and CI is green — ready for another look when you have a moment.

Since the rebase, three review rounds landed (commits `6a6089a0`, `5137c6d9`, `bd726e24`, `a71f19a5`):

**Round 1 — your review notes**
- ODPS product re-import round-trip (routes non-snake_case payloads through `parse_odcs`)
- UC additional-domain tags emitted separately — numbered keys `data_domain_1/_2/…` (a UC securable holds one value per key, so they can't share `data_domain_additional`; primary stays `data_domain`)
- `list_products` N+1 batched
- Marketplace domain filter is now multi-select (any-of union)

**Round 2**
- Data-product list route applied the domain filter after the default `limit=100` → under-reported; now loads the full candidate set when filtering
- `teams_manager` now clamps an out-of-set `primary_domain_id` (was surfacing as HTTP 500)
- UC numbered keys skip any already-used key to avoid clobbering an unrelated tag

**Round 3 (most important)**
- **Export round-trip data loss:** both the ODCS (contracts) and ODPS (products) builders rebuilt `customProperties` from stored props *after* `apply_odcs` injected `additionalDomains`, overwriting it — so an entity with custom properties + 2+ domains lost its additional domains on export/re-import. Added `DomainExportAdapter.merge_custom_properties` and applied it at both sites.
- Swept two readers the column-drop missed: `compliance_entities` (was `AttributeError` on the dropped `DataProductDb.domain`) and `genie_client` (silently dropped the Domain line) — both now read the primary domain from the junction.
- Asset CSV export N+1 batched.

Backend unit suite **1308 passed / 0 failed**; frontend type-check clean; all required CI checks green.

Follow-ups filed as separate issues to keep this PR focused: #658 (search-index reindex batching), #659 (per-entity integration + dialog-payload tests), #660 (pre-existing `list_data_products` tool bug on `development`).

mvkonchits-db:
**LGTM — approving.** Re-reviewed the delta since the rebase and every note we left is resolved in code, not just in the replies. Spot-checks against the head commit (`a71f19a`):

- **ONT-NEG-002 duplicate-name guard** (the merge-conflict resolution I flagged as needing real code, not "accept both"): correctly re-expressed against the junction — resolves each name match's primary via `entity_domain_repo.get_primary_domain_id` and compares; unassigned collapses to `None == None`, preserving the original "global scope" block. No dangling `DataContractDb.domain_id` reads remain.
- All three merge conflicts clean: `_query_contracts` takes both `status=` and `domain_ids=`; `assets_manager` keeps both imports; `test-coverage.yml` on `development`'s merge-queue version with `develop` re-added.
- Export round-trip data loss fixed at both sites via `DomainExportAdapter.merge_custom_properties` (ODCS + ODPS), with tests; dropped-column readers in `compliance_entities` and `genie_client` swept onto the junction.
- **Alembic:** rebuilt the migration DAG from all 64 revisions — `l1` is the single true head, all revisions are ancestors, no drift. Single-Head Check green; migration exercised on real Lakebase Postgres 16. All required CI checks pass.

**One caveat to flag for downstream, not a blocker:** this is breaking by design (per #520) — it drops the four legacy columns and switches the REST API to `domain_ids`/`primary_domain_id` with no transition window, and UC tags move from a single `data_domain` to `data_domain` + numbered `data_domain_1/_2/…`. The MCP tool layer was migrated in lockstep and keeps stable output field names, so MCP consumers are unaffected; the break only touches callers of the raw REST endpoints and parsers of the ODCS/ODPS exports and UC tags. Worth a heads-up to anyone with a fork or integration reading the old fields before this rolls out.

_Reviewed by Isaac._

mvkonchits-db (approved):

PR #598, opened 2026-07-08
The role based data product list populating in dp tables

rgvkulkarni-dev:
The role based data product list populating in dp tables


will-yuponce-db (changes requested):
Please rebase this off of the Databricks Labs repo and not off a forked repo

PR #599, opened 2026-07-09
fix(manifest): remove iam scopes; bump to v1.0.1

mvkonchits-db:
## Summary

- Removes `iam.current-user:read` and `iam.access-control:read` from `manifest.yaml` `user_api_scopes`
- Bumps internal version from 1.0.0 → 1.0.1

## Why

These scopes were added in PR #351 with the assumption that new Marketplace installs would pick them up automatically. That assumption is wrong: the Marketplace listing manifest is fixed at publish time. Scope additions in `manifest.yaml` do not propagate to any installs — new or existing — without publishing a new listing version.

Additionally, these scope names are rejected as invalid on several Databricks workspaces, which would break any direct-deploy setup that attempts to include them via `databricks apps update`.

The diagnostic logging changes from PR #351 (authorization.py) are kept — they remain useful for triage.

## For customers who need group resolution

Direct-deploy installs can add the scopes manually via the App config UI ("User authorization" settings) if their workspace supports them. Do not include them in the default manifest until a new Marketplace listing version is published with the correct scope names confirmed against the target API.

## Test plan

- [ ] Verify `manifest.yaml` no longer contains `iam.*` entries
- [ ] Verify version reads `1.0.1` in pyproject.toml, package.json (both), and `__init__.py`
- [ ] Deploy to a test workspace and confirm app starts without scope errors


mvkonchits-db:
Follow-up context from the team: `iam.current-user:read` and `iam.access-control:read` appear to be included in the OBO token by default for all Databricks Apps (direct-deploy and git-source), so declaring them in `manifest.yaml` was redundant for those deployment types. This further confirms the removal is safe for all current customers.

The open question is whether Marketplace-distributed apps enforce a stricter allow-list where only explicitly declared scopes are forwarded. If they do, these scopes would need to be added in a new listing version with their names validated against the target API (they were rejected as invalid on at least one AWS workspace via `apps update`). That is a separate work item for the next Marketplace publish, independent of this patch.

PR #604, opened 2026-07-13
ci: fix development branch filters and back-merge runner IP failure

will-yuponce-db:
## Why

PRs targeting `development` are stuck: they never receive the five required status checks, so they stay BLOCKED forever and can never enter the merge queue. Two root causes:

1. **Stale branch filters in `test-coverage.yml`**: the `push` and `pull_request` triggers listened on `develop`, a branch that does not exist. The working branch is `development`, so PRs based on it (e.g. #598) got zero CI runs while the ruleset kept waiting on checks that would never report.
2. **Back-merge workflow fails on every push to `main`**: the job ran on stock `ubuntu-latest` runners, whose IPs are not on the org IP allow list. Every GitHub API call died with `your IP address is not permitted to access this resource`, so fixes on `main` (including the `merge_group` trigger this repo already added there) never flowed back into `development`.

## What

- `test-coverage.yml`: branch filters `[main, develop]` -> `[main, development]` on both `push` and `pull_request`; updated the Codecov comment to match.
- `backmerge-main-to-development.yml`: moved the job to the `databrickslabs-protected-runner-group` (same as every other job in this repo, and its egress IPs are on the org allow list), and rewrote the PR step to use plain REST via `gh api` instead of the GraphQL-backed `gh pr` subcommands.

## Follow-ups (not in this PR)

- `development`'s copy of `test-coverage.yml` also lacks the `merge_group` trigger; once this merges, the fixed back-merge workflow should carry it over. If the back-merge PR cannot pass checks for the same chicken-and-egg reason, an admin bypass merge onto `development` is needed once.
- #577 proposed deleting the back-merge workflow as non-functional; with this fix it works, so #577 can likely be closed.

### DIFF
--- a/.github/workflows/backmerge-main-to-development.yml
+++ b/.github/workflows/backmerge-main-to-development.yml
@@ -21,7 +21,13 @@ concurrency:
 
 jobs:
   backmerge:
-    runs-on: ubuntu-latest
+    # Must run on the protected runner group: the databrickslabs org has an IP
+    # allow list, and stock GitHub-hosted runner IPs are not on it, so any
+    # GitHub API call from ubuntu-latest fails with "your IP address is not
+    # permitted to access this resource".
+    runs-on:
+      group: databrickslabs-protected-runner-group
+      labels: linux-ubuntu-latest
     steps:
       - name: Checkout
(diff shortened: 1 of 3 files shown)


larsgeorge-db (approved):
lgtm

PR #617, opened 2026-07-17
ci: back-merge with a merge commit so the sync stops replaying

will-yuponce-db:
## Problem

PR #616 (and #615 before it) re-apply main's entire history onto `development` even though the content is already there. Root cause is the **merge method**, not the workflow's PR logic:

- The `development` ruleset (`main + development protection and merge queue`, id 17892460) allows **only the `squash` merge method**, and the merge queue also squashes.
- Squash-merging a `main -> development` PR collapses all of main's commits into **one brand-new commit that lives only on `development`** and carries none of main's SHAs (confirmed: #615's merge commit `bc0e502b` has a single parent and is not on `main`).
- So the **merge-base of `main` and `development` never advances** (still `cba9fcc6`). The workflow's guard `git merge-base --is-ancestor origin/main origin/development` stays false, and every push to `main` regenerates a PR replaying the whole history (`development...main` = 41 commits / ~6.5k insertions), while the *actual* content delta (`development..main`) is only ~7 lockfile/CI files.

## Fix

Make the workflow push a **real merge commit** (`git merge --no-ff`, with `main` as a second parent) straight to `development`. A merge commit advances the merge-base, so the next run's ancestor check short-circuits and nothing is reopened — the sync becomes idempotent.

## Required follow-up (repo admin)

Direct pushes to `development` bypass the PR flow, so the pushing token **must be added to the ruleset's `bypass_actors`** (it currently has none) and supplied as the **`BACKMERGE_TOKEN`** secret. If that secret is absent, or the merge hits conflicts, the workflow **falls back to opening a PR** for manual (merge-commit) resolution instead of force-pushing — so this change is safe to merge before the token/bypass is configured; it just won't stop the loop until then.

Note: this touches `contents: write` (needed to push the merge commit).


mvkonchits-db (commented):
The workflow change itself is correct and the root-cause diagnosis holds up — I confirmed #615's back-merge commit `bc0e502b` has a single parent and is not on `main`, so the squash produced a dev-only commit and the merge-base never advanced. The `--no-ff` merge-commit fix genuinely advances the merge-base and makes the sync idempotent, and the fallback-to-PR path (no `BACKMERGE_TOKEN` or a conflicting merge) is a sound safety design. `contents: write` is correctly added.

**The blocker isn't the conflicts — it's the base branch.** `ci/backmerge-merge-commit` was cut from `development`, not `main`, but it targets `main`:

- merge-base(head, `main`) = `cba9fcc6` (the stale base)
- head is effectively `development` + 1 commit

That's why the diff shows 103 files / +6759 / -675. Only **one file is actually yours** (`backmerge-main-to-development.yml`, +70/-19 vs main's current version). The other ~102 are `development`-ahead-of-`main` drift — identical between the head branch and `development`. `main` is currently 52 ahead / 29 behind `development`.

Consequence: merging this into `main` wouldn't just add the workflow, it would forward-merge ~29 `development` commits into the release branch. The reported conflicts are a symptom of the same wrong-base problem, so resolving them still leaves you merging all of `development` into `main`.

Suggest re-cutting from `main` so the PR is just the workflow file:

```bash
git fetch origin
git checkout -B ci/backmerge-merge-commit-v2 origin/main
git checkout origin/ci/backmerge-merge-commit -- .github/workflows/backmerge-main-to-development.yml
git commit -m "ci: back-merge with a merge commit so the sync stops replaying"
git push origin ci/backmerge-merge-commit-v2
```

That should give a 1-file, zero-conflict PR.

Minor notes for the clean version:
- The `test-coverage.yml` `merge_group` + `develop`->`development` change is genuinely useful (queued PRs hang without a `merge_group` trigger), but right now it's riding along as dev-drift rather than part of your commit. Worth deciding whether to include it explicitly here or let it land via the normal sync.
- The new workflow pins `actions/checkout@df4c...` (v6) while the drifted `test-coverage.yml` moves to v7 — worth keeping consistent.
- The repo-admin follow-up (token in the ruleset `bypass_actors` + `BACKMERGE_TOKEN` secret) is the actual gating item for stopping the loop; the PR is upfront that it's a no-op on the loop until then.

mvkonchits-db (approved):
Re-reviewed after the re-cut. The wrong-base problem is fixed: `main` (`01375ee`) is now an exact ancestor of the head branch, the diff dropped from 103 files to 10 (+91/-40), `mergeable` is clean, and all required checks pass (Backend, Frontend, TS, Alembic single-head, lockfile). The workflow change itself is unchanged and still correct, and the `test-coverage.yml` `merge_group` + `develop`->`development` fix I flagged earlier is now already on `main`, so it's out of the diff. Approving on that basis.

One thing to decide before merge (non-blocking): the re-cut pulled in 5 dependabot bumps on top of the workflow change, and this PR now forward-ports them to `main`:

- `reportlab` 4.4.10 -> 5.0.0 (a major bump)
- `pydantic-settings` 2.14.1 -> 2.14.2
- `@radix-ui/react-popover` 1.1.15 -> 1.1.19
- `actions/checkout` 6 -> 7
- `astral-sh/setup-uv` 8.2.0 -> 8.3.2

I checked the one that matters. `reportlab` 5.0.0's only breaking change is that `trustedHosts=None` now trusts no hosts in `open_for_read` (a remote-asset loading restriction). The single consumer here, `src/backend/src/common/agreement_pdf.py`, uses only the local `pdfgen` canvas surface (`Canvas`, `pagesizes.letter`, `units.inch`, `setFont`, `drawString`, `showPage`) with no remote assets, so that change doesn't touch this path. No Python versions were dropped, and the app's `>=3.11,<3.13` pin is still fine. It's covered by three unit tests and Backend Tests are green.

So the runtime risk is low. The real concern is provenance: a PR titled "ci: back-merge..." is quietly shipping a major runtime dependency to the release branch where a reviewer approving a CI fix wouldn't scrutinize it. Two clean options:

1. Keep the bumps and add a line to the PR body naming them (e.g. "also forward-ports dependabot #566/#565/#560/#559/#612"), or
2. Strip them with `git rebase --onto origin/main <workflow-commit>^ <workflow-commit>` for a pure workflow-only PR and let the normal sync carry the deps.

I'd lean toward stripping so the reportlab major lands on its own reviewable line, but either is fine given CI is green. Approving so this isn't blocked on it.

PR #558, opened 2026-06-22
chore: bump version to 1.0.0

will-yuponce-db:
## v1.0.0 release

Version bump for the **V1.0.0** release cut.

- Bumps version 0.6.1 → 1.0.0 across `src/pyproject.toml`, `src/backend/src/__init__.py`, `src/frontend/package.json`, `src/package.json` (via `src/scripts/bump_version.py`).
- Regenerates `src/docs/RELEASE_NOTES.md` from `v0.6.1` (via `src/scripts/generate_release_notes.py`), covering all features/fixes/maintenance since 0.6.1 — including the V1-hide work (#553), the data-products fixes (#549/#554/#556), ontology fixes (#540), and #557.

Once merged, the `v1.0.0` tag will be re-pointed to the squashed commit on `main` and a GitHub Release published.

PR #618, opened 2026-07-20
fix(build): correct ruff target-version and harden bump_version.py version replacement

surojitchowdhury:
## Summary

`[tool.ruff]` in `src/pyproject.toml` had `target-version = "1.0.1"` — the project's application version, which is **not** a valid ruff target. Ruff only accepts language targets (`py37`..`py315`), so it fails at **config-parse time** before linting anything:

```
ruff failed
  Cause: Failed to parse .../src/pyproject.toml
  Cause: TOML parse error at line 128, column 18
128 | target-version = "1.0.1"
unknown variant `1.0.1`, expected one of `py37`, `py38`, ... `py315`
```

This breaks `hatch -e dev run lint`, `ruff format --check`, and the ruff pre-commit hooks for every contributor.

This resolves the semantics issue a reviewer flagged during #558: *"target-version seems to be referring to the Python language version like py311."*

## What changed

- **`src/pyproject.toml`** — set `[tool.ruff] target-version = "py311"` (matches `requires-python = ">=3.11,<3.13"`).
- **`src/scripts/bump_version.py`** — fixed the **root cause**. The pyproject version-bump used an unanchored regex (`(version\s*=\s*")[^"]+(")`) with a global `re.sub`, which also matched `target-version` — so every version bump rewrote ruff's `target-version` with the app version. The replacement is now scoped to the `[project]` table (`(?ms)(^\[project\](?:(?!^\[).)*?^version\s*=\s*")[^"]+(")`) with `count=1`, so it can only ever touch the project version.
- **`src/backend/tests/test_bump_version_scoping.py`** — new hermetic regression test proving a bump rewrites **only** `[project].version` and leaves `target-version` and any other table's `version` key untouched. Fails against the old unanchored behavior.

## Why

Without this, backend lint/format and the ruff pre-commit hooks are unusable out of the box, and the next `bump_version.py` run would silently re-introduce the invalid value.

## How to test

```bash
# ruff config now parses and runs (previously errored at parse time)
cd src && hatch -e dev run lint
cd src && hatch -e dev run ruff format backend --check

# regression test for the bump_version scoping fix
cd src && hatch -e dev run python -m pytest backend/tests/test_bump_version_scoping.py -v
# -> 3 passed

# full suites
cd src && hatch -e dev run test            # 1452 passed
cd src/frontend && yarn type-check && yarn test:run
```

## Notes

- Scope is intentionally limited to fixing the broken config + its root cause. Now that ruff's config parses, `ruff check`/`format` surface substantial **pre-existing** lint/format debt in `backend/` — that cleanup is deliberately **out of scope** for this fix and left for a follow-up.
- CI is unaffected: no workflow currently invokes ruff.

### DIFF
--- a/src/backend/tests/test_bump_version_scoping.py
+++ b/src/backend/tests/test_bump_version_scoping.py
@@ -1,0 +1,102 @@
+"""Regression tests for the ``scripts/bump_version.py`` pyproject.toml scoping fix.
+
+The version bumper edits ``pyproject.toml`` with a regex substitution. A prior
+bug used an unanchored pattern with a global ``re.sub``, so bumping the project
+version also clobbered ``[tool.ruff]``'s ``target-version`` (and would clobber a
+``version`` key under any other table). The fix scopes the pattern to the
+``[project]`` table and passes ``count=1``.
+
+These tests pin that behavioral guarantee: bumping rewrites ONLY the
+``[project]`` table's ``version`` key and leaves ``target-version`` and a decoy
+table's ``version`` untouched. They are hermetic — the module is loaded straight
+from its file path and all edits happen against ``tmp_path``, never the real
+repo ``pyproject.toml``.
+"""
+from __future__ import annotations
+
+import importlib.util
+from pathlib import Path
+
+import pytest
+
+# Load scripts/bump_version.py directly by path so the test does not depend on
+# ``scripts`` being importable as a package from the pytest rootdir.
+_SRC_ROOT = Path(__file__).resolve().parents[2]  # .../src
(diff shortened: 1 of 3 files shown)


surojitchowdhury:
Closing as superseded: the ruff `target-version` bug this PR addressed (originally flagged in #558) has already been fixed on `main` — `target-version` is now `py311` and `scripts/bump_version.py` was anchored to only match the top-level project `version` key (landed with the v1.0.2 release, #661). Since the fix is already in, there's nothing left for this PR to resolve. Thanks!

PR #630, opened 2026-07-21
fix: support local Lakebase OAuth users

kremerlal:
Allow Databricks email roles and require SSL so local OAuth connections authenticate correctly.


kremerlal:
meant to push to my own branch

PR #635, opened 2026-07-22
fix(mcp): comply with JSON-RPC and release DB session for streaming clients

nehmetohmedb:
## Problem

Two transport-layer defects broke MCP clients that reconnect or validate strictly (Kasal / MCP Python SDK / CrewAI):

1. **DB session pinned for the life of an SSE stream.** FastAPI holds route dependencies until the response finishes; for the GET SSE stream that pinned one pooled DB connection per open stream for the stream's whole lifetime. Auto-reconnecting clients exhausted the pool within seconds, after which token lookups failed (`401 'Invalid or missing API key'`) and other requests hung on the pool timeout. The stream loop only uses the in-memory session store, so the DB session can be returned up front.
2. **JSON-RPC spec violations.** Unknown notifications (e.g. `notifications/cancelled` from CrewAI) got a `METHOD_NOT_FOUND` error with `id=null`, which the MCP Python SDK cannot parse (SSE stream crashed, reconnect loop). `model_dump()` emitted both `result` and `error` keys (one null), rejected by strict `JSONRPCMessage` validation. The body was parsed after auth, so auth/validation errors echoed `id=null`.

## Fix

- Release the DB session before entering the long-lived SSE loop.
- Never reply to notifications (return `202 Accepted`, no body, per the streamable HTTP spec); serialize exactly one of `result`/`error`; parse the body before auth so errors echo the request id.

## Testing

- Backend unit suite: 1449 passed, 1 skipped (matches `main` baseline, no regressions).
- Verified on a live deployment: MCP Python SDK / Kasal client connects, lists tools, and calls tools without the reconnect loop or pool exhaustion.

### DIFF
--- a/src/backend/src/routes/mcp_routes.py
+++ b/src/backend/src/routes/mcp_routes.py
@@ -495,6 +495,15 @@ async def mcp_sse_stream(
         }
         logger.info(f"Created new MCP session: {session_id}")
     
+    # Release the DB connection BEFORE streaming: FastAPI keeps dependencies
+    # alive until the response completes, which for a long-lived SSE stream
+    # means each open stream pins one pooled connection indefinitely.
+    # Reconnecting clients (Kasal retries on any hiccup) exhaust the pool
+    # within seconds — after which token validation can't query and every
+    # request 401s or hangs on pool timeout. The stream loop only touches
+    # the in-memory session dict, so the session is safe to close here.
+    db.close()
+
     # Return SSE stream
     return EventSourceResponse(
         sse_stream_generator(token_info, session_id, request),
--- a/src/backend/src/routes/mcp_routes.py
+++ b/src/backend/src/routes/mcp_routes.py
@@ -78,6 +78,20 @@ def make_success_response(
     return JSONRPCResponse(result=result, id=request_id)
 
 
+def dump_jsonrpc_response(resp: JSONRPCResponse) -> Dict[str, Any]:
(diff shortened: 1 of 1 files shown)


nehmetohmedb:
### A note on the failing CI checks

All non-title checks fail within ~10s at the very first step (`.github/actions/setup-node-jfrog`) with:

```
ACTIONS_ID_TOKEN_REQUEST_TOKEN: unbound variable
```

This is the standard GitHub restriction for PRs opened from a fork: the OIDC id-token (and repo secrets) needed to authenticate to Databricks' private JFrog registry are not exposed to fork-triggered runs, so every job exits before installing dependencies or running any test. A maintainer approving the workflow runs (or re-running them with base-repo context) should clear this.

The branch was validated locally against the latest `main` (`7b4fb65a`):

- **Backend unit suite: 1449 passed, 1 skipped** — identical to the `main` baseline on the same machine (no regressions).
- The behavioral fix was additionally verified against a live Ontos deployment (details in the PR description).

Happy to rebase or adjust anything on request.

nehmetohmedb:
Superseded by #642, which carries the same fix as a branch inside `databrickslabs/ontos` so CI runs in a trusted context (all checks green). Closing this fork-based duplicate; the checks here only failed at the `setup-*-jfrog` OIDC step, a fork-PR limitation, never at any code check. Please review #642 instead.

PR #640, opened 2026-07-24
fix(compliance): app-entity loader yields nothing and repo is fail-closed

nehmetohmedb:


mvkonchits-db (approved):
Approved. Verified against main: repositories expose get_multi (not list, which has 0 defs), and the old import modules (domains_repository, data_asset_review_repository) don't exist, so the loader was fully broken. Per-type try/except is a genuine improvement over the single blanket catch.

Two non-blocking follow-ups:
- Every entity type loads get_multi(limit=1000). Estates with >1000 products/contracts silently under-report compliance coverage with no truncation log. Consider a paginate-until-exhausted loop, or at minimum a warning when a type hits the cap.
- The per-type 'log and continue' now means compliance can run on partial data; pairs with the cap above.

PR #641, opened 2026-07-24
fix(mcp): scoped tokens cannot list or call read-scoped tools

nehmetohmedb:


mvkonchits-db (approved):
Approved. Verified against main: _has_scope does ':' in required_scope with no None guard, and app_state/handbook tools set required_scope=None, so tools/list crashed for any non-wildcard token. Guard + explicit scopes on the CRUD tools are correct.

PR #642, opened 2026-07-24
fix(mcp): comply with JSON-RPC and release DB session for streaming clients

nehmetohmedb:


mvkonchits-db (approved):
Approved. All three sub-fixes verified as real: JSON-RPC single-key serialization (strict clients reject the dual null key), db.close() before the SSE stream (safe — the generator re-acquires its own session via Depends), and 202 for notifications.

One nuance for the record (not blocking): the notifications/* early-return now runs BEFORE validate_api_key, so notifications/initialized mutates _sessions[session_id]['initialized'] on an unauthenticated request. Exploitability is low since session IDs are secrets.token_urlsafe(32), but on main this state change sat behind auth. If you'd prefer, move the auth check above the notification branch (still returning 202 after validating the key). Approving as intentional per the transport-compliance goal.

PR #643, opened 2026-07-24
fix(mcp-tools): make agent data access actually work

nehmetohmedb:
- app.state.workspace_client was never set, so the analytics MCP tools (get_table_schema, execute_analytics_query, explore_catalog_schema) always failed with 'Workspace client not available' — agents could not query any data through Ontos.
- get_data_product ignored Description model objects (description came back null) and returned no output ports — agents saw a bare product with no pointers to the physical tables.
- search_data_contracts read DataContractsManager._contracts, a legacy in-memory dict that is never populated — always zero results. Query the contracts table directly.

Co-authored-by: Isaac
(cherry picked from commit cd809ac25ef1b5b080b3bbd50c402ead30682236)

### DIFF
--- a/src/backend/src/app.py
+++ b/src/backend/src/app.py
@@ -179,6 +179,11 @@ async def startup_event():
         from src.controller.grant_manager import init_grant_manager
         from src.common.workspace_client import get_workspace_client
         ws_client = get_workspace_client(settings=settings)
+        # Expose the app-identity workspace client for MCP analytics tools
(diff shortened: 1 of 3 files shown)


mvkonchits-db (approved):
Approved. Verified against main: ToolContext.workspace_client was never populated at startup, and ctx.db.query(...) is the established pattern across the tools. Fix is correct.

PR #644, opened 2026-07-24
fix(ontology-generator): retry without temperature for Claude endpoints

nehmetohmedb:
Databricks foundation-model endpoints for Claude (e.g. us.anthropic.claude-opus-4-8) reject the temperature parameter with 400 BAD_REQUEST, so generation always failed on those endpoints. Retry the request without temperature when the endpoint rejects it.

Co-authored-by: Isaac
(cherry picked from commit c65dfd50dac69ea4f7d63584aaab3222d9ae8652)


mvkonchits-db (approved):
Approved. Verified against main: temperature=0.1 was hard-coded in both the primary and fallback calls; Databricks Claude endpoints reject it with 400. Retry-without-temperature logic is sound.

PR #655, opened 2026-07-28
Fix non-deterministic frontend build on Databricks Apps (npm ERESOLVE)

### LINKED ISSUE #654: [BUG]: Databricks Apps deploy fails — frontend npm install ERESOLVE (non-deterministic build without lockfile)

### Current Behavior

Deploying Ontos as a Databricks App (including via the Marketplace listing) fails during the frontend build. The backend installs fine, then `scripts/build_static.sh` runs and the deploy fails within ~19s of "Building frontend...", before Vite compiles anything.

Root cause: the frontend build is non-deterministic in the Databricks Apps runtime. `src/frontend/.npmrc` sets `package-lock=false` and there is no `package-lock.json` (only `yarn.lock`). `build_static.sh` prefers yarn but falls back to npm when yarn is absent, which is the case in the Databricks Apps build image. So it runs `npm ci` (fails, no lockfile) → `npm install`, which resolves every `^`-ranged dependency to the newest version on the registry at build time. It recently started resolving `@hookform/resolvers` to `5.5.7`, whose `peerOptional ajv-formats@^2` conflicts with the project's `ajv-formats@^3`, so npm aborts with `ERESOLVE`. The same source built fine a few weeks ago because the registry served an older, compatible version. No source change is involved.

### Expected Behavior

`databricks apps deploy` / the Marketplace listing builds the frontend reliably and reproducibly, independent of what has since been published to the npm registry.

### Steps To Reproduce

1. In an environment without yarn (as the Databricks Apps build image), from `src/frontend`.
2. `npm ci` → fails (no `package-lock.json`).
3. `npm install` → `ERESOLVE`: `@hookform/resolvers@5.5.7` peer `ajv-formats@^2` vs project `ajv-formats@^3`.

### Proposed fix

- Quick unblock: add `legacy-peer-deps=true` to `src/frontend/.npmrc` (verified: install + `vite build` then succeed).
- Robust: make the build deterministic in the npm path, either by committing a `package-lock.json` or by ensuring yarn (with the existing `yarn.lock` and `--frozen-lockfile`) is actually used in the Databricks Apps build.

### Cloud

GCP

### Browser

Chrome

### Version

latest

### Relevant log output

```shell
Building frontend in /app/python/source_code/frontend...
npm error code ERESOLVE
npm error While resolving: @hookform/resolvers@5.5.7
npm error Found: ajv-formats@3.0.1
npm error Could not resolve dependency:
npm error peerOptional ajv-formats@"^2.1.1" from @hookform/resolvers@5.5.7
npm error Conflicting peer dependency: ajv-formats@2.1.1
npm error Fix the upstream dependency conflict, or retry
npm error this command with --force or --legacy-peer-deps
```

lucas-M-Bernardo:
Fixes #654.

## Problem

Deploying Ontos as a Databricks App fails during the frontend build. In the Databricks Apps build image there is no yarn, so `scripts/build_static.sh` falls back to the npm path: `npm ci` fails (no `package-lock.json`, since `.npmrc` sets `package-lock=false`), then `npm install` resolves every `^`-ranged dependency to the newest version on the registry. It recently began resolving `@hookform/resolvers@5.5.7`, whose `peerOptional ajv-formats@^2` conflicts with the project's `ajv-formats@^3`, so npm aborts with `ERESOLVE` before Vite runs. The same source built fine weeks ago, no source change involved.

## Fix

Add `legacy-peer-deps=true` to `src/frontend/.npmrc` so the npm fallback resolves peer deps instead of aborting. Verified locally (Node 24/25, npm 11, no yarn): `npm install` + `vite build` succeed and emit `static/`.

## Note

This unblocks the build but the npm path stays non-deterministic. A more robust follow-up would be committing a `package-lock.json`, or ensuring yarn (with the existing `yarn.lock` and `--frozen-lockfile`) is actually used in the Databricks Apps build.

### DIFF
--- a/src/frontend/.npmrc
+++ b/src/frontend/.npmrc
@@ -1,1 +1,2 @@
 package-lock=false
+legacy-peer-deps=true


larsgeorge-db:
This seems resolved now witht he switch to pure `npm` building. If not, please upate or kindly submit a new PR.

PR #673, opened 2026-08-05
fix(search,demo): correct asset search link + invalid workflow step_type in demo seeds

mvkonchits-db:
## Summary

Two related fixes surfaced while validating the demo data end to end on a fresh deploy.

### 1. Asset global-search results 404
`AssetsManager` built search-result links as `/governance/assets/{id}`, but that route does not exist in the frontend router — only `/assets/:assetId` is registered (`src/frontend/src/app.tsx`). Every asset hit from global search therefore 404s. Fixed the link to `/assets/{id}`.

### 2. Demo seeds insert an invalid workflow `step_type`, crashing `/api/workflows`
The `auto`, `fsi`, `hls`, and `mfg` demo packs insert `workflow_steps` with `step_type='manual_approval'`, which is not a member of the `StepType` enum. Once any of these presets is loaded, `GET /api/workflows` returns **500** for the whole list:

```
ValueError: 'manual_approval' is not a valid StepType
  at workflows_manager.py  step_type=StepType(step.step_type)
```

The same steps also used the dead config key `approver_role` with a bare business-role UUID. The approval step handler reads `config['approvers']`, and the role resolver requires the `business:` prefix to resolve a `BusinessRoleDb` id — so even after the enum fix the steps would fail at execution with "No approvers configured". Corrected both:
- `step_type`: `manual_approval` → `approval`
- `config`: `{"approver_role": "<uuid>"}` → `{"approvers": "business:<uuid>"}`

## Testing
- Deployed to a dev workspace, loaded `preset=auto`.
- Before: `GET /api/workflows` → 500; asset search click → 404.
- After: `GET /api/workflows` → 200; seeded approval workflow shows `approvers: business:<uuid>`; asset search click opens the asset detail page.

## Notes
- Only `retail` was already clean; the other four packs are fixed here.
- Observed (out of scope, not fixed here): `DELETE /api/settings/demo-data` skips one `assets` delete (`0f9%`) due to a FK constraint on `asset_relationships` — pre-existing, unrelated to these changes.

### DIFF
--- a/src/backend/src/controller/assets_manager.py
+++ b/src/backend/src/controller/assets_manager.py
@@ -617,7 +617,7 @@ class AssetsManager(SearchableAsset):
             type=f"asset-{type_name.lower().replace(' ', '-')}",
             title=asset_db_obj.name,
             description=asset_db_obj.description or '',
-            link=f"/governance/assets/{asset_db_obj.id}",
+            link=f"/assets/{asset_db_obj.id}",
             tags=tags,
             feature_id="assets",
             extra_data={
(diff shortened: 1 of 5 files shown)


larsgeorge-db (approved):
lgtm

PR #674, opened 2026-08-06
feat(dqx): gate Profile with DQX button on job state + optional admin request

larsgeorge-db:
## Problem

Users clicked **Profile with DQX** while the backing background job was disabled and got a generic error toast. The button gave no indication that the feature was unavailable, and there was no path to ask an admin to turn it on.

## Approach

The button now reflects the job's install state, explains why it is disabled, and can optionally notify admins.

The main constraint: every existing endpoint that exposes job state (`/api/settings`, `/api/jobs/workflows/statuses`) is gated on `PermissionChecker('jobs', READ_ONLY)`, and `jobs` is absent from `DEFAULT_ROLE_PERMISSIONS` for **all** non-admin roles. So exactly the users who see this button (Data Steward / Data Producer, who have `data-contracts` READ_WRITE) would get a 403 from all of them. Hence a new minimal capability endpoint rather than reusing what exists.

### Backend

- **`ALLOW_JOB_ENABLEMENT_REQUESTS` setting** — persisted via `app_settings`, mirroring the existing `LLM_ENABLED` load/get/save pattern. Controls whether users may ask admins to enable a background job. Defaults to `false`.
- **`GET /api/jobs/capabilities`** — authenticated-only, no `jobs` permission required. Reports install state for an allowlist of feature-gating workflows (`dqx_profile_datasets`, `mdm_match_detect`). Returns booleans only — deliberately no job IDs, cluster IDs or run history, so it cannot become a job inventory for unprivileged users.
- **`POST /api/jobs/workflows/{id}/request-enablement`** — sends a deduplicated `ACTION_REQUIRED` notification to the `Admin` role, following the existing access-grant / deploy-request pattern. Not gated by the `jobs` permission by design: the point is to let users who *cannot* manage jobs ask someone who can. Guarded instead by the setting and the workflow allowlist.
- **`NotificationsManager.has_unhandled_actionable_notification()`** — dedupe helper so repeated clicks don't flood the admin bell. Fails open: a duplicate notification is preferable to silently dropping a request.

### Two bugs found while wiring this up

Both in the profiling start path, both user-visible:

1. **Dangling profiling runs.** `start_profiling()` committed the run record *before* checking the workflow was installed, so every failed click left a permanent `pending` row. The UI reads the latest run, sees `pending`, and shows *"DQX profiling in progress..."* indefinitely — with no job in existence to ever advance it. The check now runs first.
2. **Swallowed error message.** The route collapsed the sanitized `ValueError` into `detail="Invalid profiling request"`, which is why the toast was useless. It now raises `ConflictError` (409) with an actionable message and re-raises `HTTPException` so it reaches the client. (`.cursor/rules/05-code-style-and-structure.mdc:28` explicitly permits exposing sanitized `ValueError` messages.)

### Frontend

- **`job-capabilities-store`** — assumes the job is available until proven otherwise, so a failed capabilities fetch can never disable a working feature.
- **Button + hint** — disabled when the job isn't installed, with the hint in a tooltip and the *Notify admin* action **inside** the tooltip, so the request doesn't compete visually with the primary schema actions.
- **Stale-run guard** — a lingering `pending`/`running` record is treated as stale when the job isn't installed, fixing the false in-progress spinner.
- **Settings > General** toggle for admins, with translations for all 7 locales.

## Verification

Exercised end-to-end against a local backend + frontend on isolated ports:

| Check | Result |
|---|---|
| `GET /api/jobs/capabilities` | correct booleans, allowlist only |
| `POST .../request-enablement`, flag off | `403` |
| `POST .../request-enablement`, flag on | `{"requested": true}` |
| repeat request | `{"requested": false, "reason": "already_requested"}` |
| non-allowlisted workflow ID | `404 Unknown workflow` |
| admin notification | lands with resolved display name, correct payload |
| setting round-trip | persists through `PUT`/`GET /api/settings` |
| disabled button + tooltip | renders; in-tooltip *Notify admin* click fires |
| false in-progress spinner | gone |

`tsc --noEmit` clean. No new lint errors — `data-contract-details.tsx` has 24, verified identical to its pre-change baseline.

## Follow-ups (not in this PR)

- **Existing deployments may hold dangling `pending` profiling runs** created before this fix. The UI now ignores them, but a cleanup `UPDATE` or migration would remove them properly.
- **Pre-existing, unrelated to this change:** `can_delete=False` is stored correctly in Postgres but the API returns `true` for every notification — confirmed against the existing `workflow_approval` notification from `workflow_executor.py:631`, so it affects all notification flows. Worth its own issue.

### DIFF
--- a/src/backend/src/common/config.py
+++ b/src/backend/src/common/config.py
@@ -158,6 +158,11 @@ class Settings(BaseSettings):
     sandbox_allowed_schemas: List[str] = Field(default_factory=lambda: ['sandbox'], validation_alias=AliasChoices('SANDBOX_ALLOWED_SCHEMAS', 'sandbox_allowed_schemas'))
     sandbox_enforce_allowlist: bool = Field(True, validation_alias=AliasChoices('SANDBOX_ENFORCE_ALLOWLIST', 'sandbox_enforce_allowlist'))
 
+    # Background job enablement requests
+    # When True, users who hit a feature backed by a disabled background job may
+    # send an ACTION_REQUIRED notification to the Admin role asking for it to be enabled.
+    ALLOW_JOB_ENABLEMENT_REQUESTS: bool = Field(False, env='ALLOW_JOB_ENABLEMENT_REQUESTS')
+
     # Delivery Mode settings
     # Controls how governance changes (GRANTs, tag assignments, etc.) are propagated
     DELIVERY_MODE_DIRECT: bool = Field(False, env='DELIVERY_MODE_DIRECT')  # Apply changes directly to Unity Catalog via SDK
--- a/src/backend/src/controller/data_contracts_manager.py
+++ b/src/backend/src/controller/data_contracts_manager.py
@@ -39,6 +39,7 @@ from src.common.search_registry import searchable_asset
 
 from src.common.logging import get_logger
 from src.common.database import get_session_factory
+from src.common.errors import ConflictError
 from src.db_models.data_contracts import (
     DataContractDb,
     DataContractTagDb,
@@ -3840,6 +3841,20 @@ class DataContractsManager(DeliveryMixin, SearchableAsset):
         if not contract:
             raise ValueError("Contract not found")
         
+        # Get jobs manager and workflow installation.
+        # Checked BEFORE the run record is created: a failed precondition used to leave
+        # behind a 'pending' run that the UI would then poll indefinitely.
+        if not jobs_manager:
+            raise ValueError("Jobs manager not available")
+
+        workflow_id = "dqx_profile_datasets"
+        installation = workflow_installation_repo.get_by_workflow_id(db=db, workflow_id=workflow_id)
+        if not installation:
+            raise ConflictError(
+                "The DQX profiling background job is not enabled. "
+                "Ask an administrator to enable it under Settings > Jobs."
+            )
+
         # Create profiling run record
         profile_run_id = str(uuid4())
         run = DataProfilingRunDb(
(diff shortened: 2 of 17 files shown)


mvkonchits-db (approved):
Reviewed the full diff and verified the backend dependencies against source (not just the description). High-quality PR — approving.

**The permission gate is the crux, and it's handled correctly.** Every existing endpoint exposing job state is gated on `PermissionChecker('jobs', READ_ONLY)`, and `jobs` isn't in `DEFAULT_ROLE_PERMISSIONS` for any non-admin role — so the exact users who see this button (Data Steward / Data Producer) would 403 on all of them. The new authenticated-only `GET /jobs/capabilities` returning booleans-only for an allowlist (`dqx_profile_datasets`, `mdm_match_detect`) is the right answer: it avoids becoming a job inventory for unprivileged users while still letting the UI reflect state.

**Verified against source:**
- `ConflictError` extends `AppError(HTTPException)`, so the route's `except HTTPException: raise` genuinely propagates the 409 with its actionable message.
- `get_all_installed`, `get_by_workflow_id`, `list_available_workflows`, `create_notification`, and `get_multi(db, *, limit=)` all exist with matching signatures; `json` is imported in `notifications_manager`.
- Both bug fixes are real and correct: the install check now runs *before* the run record is committed (no more permanent dangling `pending` rows), and `ConflictError(409)` replaces the swallowed `detail="Invalid profiling request"`.
- Secure defaults: `ALLOW_JOB_ENABLEMENT_REQUESTS` defaults `False`; the frontend store assumes *installed* until proven otherwise so a failed capabilities fetch can't disable a working feature. All 7 locales updated.

**Non-blocking observations:**
- `has_unhandled_actionable_notification` loads up to 5000 notifications and JSON-parses each in Python for the subset match. Fine at current scale and it fails open, but if notification volume grows this wants a filtered query on `action_type`/`read` rather than a full scan.
- Two rapid clicks can both pass the dedupe check before either commits, creating duplicate admin notifications. Acceptable given the fail-open design; just noting it.
- The two self-flagged follow-ups (cleanup migration for pre-existing dangling runs; the unrelated `can_delete=True` API bug) are the right ones to split out.

PR #679, opened 2026-08-07
fix(notifications): enforce can_delete and recipient access on DELETE

### LINKED ISSUE #675: fix(notifications): DELETE endpoint ignores can_delete and skips recipient access check
## Summary

`DELETE /api/notifications/{notification_id}` does not honour the `can_delete` flag, and unlike the other notification mutation endpoints it never verifies the caller is actually a recipient. Both checks exist in the codebase; this one endpoint just doesn't call them.

Found incidentally while working on #674 (unrelated change).

### 1. `can_delete` is never enforced server-side

Several flows deliberately create notifications with `can_delete=False` to mean *"you must respond to this, you cannot dismiss it"*:

- `common/workflow_executor.py:631` — approval requests (`# Must respond to this notification`)
- `controller/access_grants_manager.py:765` — access grant requests
- `controller/data_contracts_manager.py:5264` — contract deploy requests
- `routes/user_routes.py:449` — role access requests

The value is persisted correctly (verified directly against Postgres). But `manager.delete_notification()` (`controller/notifications_manager.py:283-290`) goes straight to `self._repo.remove(...)` with no check on the flag, so the constraint is enforced by the frontend alone and any direct API call bypasses it.

### 2. No recipient access check

`NotificationsManager.can_user_access_notification()` exists and is used by:

- `routes/notifications_routes.py:153` — the mark-as-read endpoint
- `routes/workflows_routes.py:1224` — the workflow approval handler

But **not** by the delete endpoint (`routes/notifications_routes.py:90-98`), which is guarded only by `PermissionChecker('notifications', FeatureAccessLevel.ADMIN)`. So any principal with that feature permission can delete any other user's notification, including ones scoped to a role they aren't in. That's inconsistent with mark-as-read, which is the weaker operation but has the stronger check.

## Reproduction

```bash
# Create/find a notification with can_delete=false, e.g. any pending approval request
psql -c "select id, title, can_delete from app_ontos.notifications where can_delete = false limit 1"

# Deletes successfully despite can_delete=false
curl -X DELETE http://localhost:8000/api/notifications/<id>
```

## Note on `can_delete` in GET responses

`get_notifications()` (`controller/notifications_manager.py:137-140`) overrides `can_delete = True` for admins:

```python
# Admins can delete any notification they can see
if is_admin:
    n.can_delete = True
```

This is **intentional** and not part of this issue — worth stating explicitly because it makes the underlying problem easy to misdiagnose: when testing as an admin, every notification reports `can_delete: true` regardless of its stored value, which masks case 1 entirely. Reproduce as a non-admin.

## Suggested fix

In `delete_notification`, mirror what the mark-as-read endpoint already does:

1. Fetch the notification and 403 if `can_user_access_notification()` returns false.
2. Reject with 403 (or 409) when `can_delete` is false on the stored row — with an explicit carve-out for admins if that matches the intent of the GET-side override above.

Worth deciding deliberately whether "admins may delete anything" should extend to `can_delete=False` rows, since those exist specifically to force a response. The GET override implies yes; the flag's purpose implies no.

## Scope

Backend only. The frontend already respects the flag it is served, so no UI change is needed for the fix itself.

ferdinand-van-butzelaar:
Closes #675.

## The two missing checks

Both already exist in the codebase; the delete endpoint just never called them.

### 1. `can_delete` was not enforced server-side

`NotificationsManager.delete_notification` went straight to `self._repo.remove(...)`. Reproduced against a session with a `can_delete=False` row present:

```
### development ###
before: can_delete=False
delete_notification returned: True
after: row DELETED

### this branch ###
before: can_delete=False
delete_notification raised: NotificationNotDeletableError: Notification … is marked as non-deletable
after: row still present
```

The four flows that rely on the flag — approval requests (`workflow_executor.py:631`), access grants (`access_grants_manager.py:765`), contract deploys (`data_contracts_manager.py:5264`), role access requests (`user_routes.py:449`) — were protected by the frontend alone.

### 2. No recipient access check

The endpoint was guarded only by `PermissionChecker('notifications', FeatureAccessLevel.ADMIN)`. That permission says the caller may use the notifications feature, not that they may act on someone else's notification. `can_user_access_notification()` was already called by mark-as-read (`notifications_routes.py:153`) and the workflow approval handler (`workflows_routes.py:1224`) — the strictly more destructive operation had the weaker check.

## The open question in the issue

> Worth deciding deliberately whether "admins may delete anything" should extend to `can_delete=False` rows, since those exist specifically to force a response. The GET override implies yes; the flag's purpose implies no.

**This PR lets admins override, and makes it explicit.** Reasoning:

- `get_notifications` already sets `can_delete = True` on everything an admin can see, and the issue confirms that override is intentional. If DELETE refused unconditionally, an admin's UI would render a delete button that always 403s — the API would report one thing and enforce another.
- There is no other cleanup path. A notification orphaned by a workflow that died mid-flight has `can_delete=False` forever and no way to respond to it.
- The override is narrow and observable: it requires the `Admin` app role, is logged at INFO, and is recorded in the audit entry as `admin_override`.

Non-admins get the strict behaviour, and `is_admin` defaults to `False` so any caller that forgets to pass it fails safe. If you'd rather have no override at all, it's a two-line change and I'm happy to make it.

## Changes

- `delete_notification(db, notification_id, *, is_admin=False)` loads the row first, raises `NotificationNotFoundError` when absent and the new `NotificationNotDeletableError` when `can_delete` is False and the caller is not an admin. A typed exception rather than a `False` return so the route can answer **403** and not **404** — the previous `if not deleted: 404` could not tell the two apart.
- The route calls `can_user_access_notification()` before deleting, mirroring mark-as-read, and returns 403 on failure.
- New `NotificationsManager.is_app_admin()` plus a `_role_maps()` helper. The role-lookup-and-detect-admin block was duplicated verbatim in `get_notifications` and `can_user_access_notification`; both now share one implementation, so what the API reports as `can_delete` and what it enforces cannot drift apart. Net −40 lines there.

Backend only. The frontend already respects the flag it is served, so no UI change is needed.

## Tests

18 new cases in `src/backend/src/tests/unit/test_notification_delete_authz.py`, against a real session rather than a mocked repository:

- `can_delete` enforcement, the admin override, and the default-to-non-admin contract
- `is_app_admin` — true for the Admin role's groups, false for a normal user, false with no user, false when the Admin role has no assigned groups, and **false when the role lookup raises** (fails closed rather than promoting on error)
- recipient access — own / other user's / role-scoped / broadcast
- route status mapping — 403 for a non-recipient, 403 for non-deletable, 404 for missing, 204 on success, and that `delete_notification` is not called at all on the two 403 paths

| | |
|---|---|
| New suite | **18 passed** |
| Full backend suite (`backend/src/tests/unit` + `backend/tests`) | **1503 passed, 1 skipped** |

The delete cases in the quarantined `test_notifications_manager.py` are updated to the new contract. That module stays quarantined — it assigns to `manager._repo.get.return_value` where `_repo` is the real repository singleton, so every case in it fails at setup, independently of this change.

## Unrelated, spotted in passing

`NotificationsManager.update_notification` is defined twice (lines 416 and 470); the second shadows the first, and ruff flags it `F811`. Same failure mode as the tool-module duplicates in #677. Not touched here — happy to file it separately.

### DIFF
--- a/src/backend/src/controller/notifications_manager.py
+++ b/src/backend/src/controller/notifications_manager.py
@@ -23,6 +23,15 @@ logger = get_logger(__name__)
 class NotificationNotFoundError(Exception):
     """Raised when a notification is not found."""
 
+class NotificationNotDeletableError(Exception):
+    """Raised when a notification is marked ``can_delete=False``.
+
+    Actionable notifications (approval requests, access grants, contract
+    deploy requests, role access requests) are created with
+    ``can_delete=False`` to mean "you must respond to this, you cannot
+    dismiss it". Only an admin may override.
+    """
+
 class NotificationsManager:
     def __init__(self, settings_manager: 'SettingsManager'):
         """Initialize the notification manager.
@@ -33,6 +42,43 @@ class NotificationsManager:
         # self.notifications: List[Notification] = [] # REMOVE In-memory list
         self._repo = notification_repo # Use the repository instance
         self._settings_manager = settings_manager # Store the manager
+
+    def _role_maps(self) -> tuple[Dict[str, Any], Dict[str, Any]]:
+        """Build (role_by_id, role_by_name) lookups for recipient resolution.
+
+        ``role_by_name`` carries the exact name plus normalized and CamelCase
+        variants so legacy string recipients still resolve. A lookup failure
+        is non-fatal and yields empty maps, which fails closed downstream.
+        """
+        role_by_id: Dict[str, Any] = {}
+        role_by_name: Dict[str, Any] = {}
+        try:
+            for role in self._settings_manager.list_app_roles():
+                role_by_id[role.id] = role
+                role_by_name[role.name] = role
+                role_by_name[role.name.lower().replace(' ', '')] = role
+                role_by_name[''.join(word.capitalize() for word in role.name.split())] = role
+        except Exception as e:
+            logger.error(f"Failed to retrieve roles for notification handling: {e}")
+        return role_by_id, role_by_name
+
+    def is_app_admin(self, user_info: Optional[UserInfo], role_by_name: Optional[Dict[str, Any]] = None) -> bool:
+        """Whether the user holds the 'Admin' app role.
+
+        This is the same notion of admin used by ``get_notifications`` when it
+        overrides ``can_delete`` to True, so what the API reports and what it
+        enforces stay in step. Distinct from ``FeatureAccessLevel.ADMIN`` on
+        the notifications feature, which is a per-feature permission.
+        """
+        if not user_info:
+            return False
+        if role_by_name is None:
+            _, role_by_name = self._role_maps()
+        admin_role = role_by_name.get('Admin')
+        if not admin_role or not admin_role.assigned_groups:
+            return False
+        user_groups = set(user_info.groups or [])
+        return any(group in user_groups for group in admin_role.assigned_groups)
 
     def get_notification_by_id(self, db: Session, notification_id: str) -> Optional[Notification]:
         """Get a single notification by ID."""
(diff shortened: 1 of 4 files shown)


mvkonchits-db (approved):
Verified every claim against `development` source rather than taking the description at face value. Clean, correctly-scoped authorization fix that closes the `can_delete` gap flagged as a follow-up in #674. Approving.

**Diagnosis confirmed:**
- `delete_notification` went straight to `_repo.remove` with no `can_delete` or recipient check (development L283–290).
- The DELETE route was guarded only by `PermissionChecker('notifications', ADMIN)` — a per-feature permission ("may use notifications"), not "may act on someone else's notification".
- mark-as-read already calls `can_user_access_notification` (routes L153); the strictly more destructive delete had the weaker check. This makes them consistent.
- `NotificationDb.can_delete` (column) and API `Notification.can_delete` (model, default True) both exist.

**Fix verified:**
- Typed exceptions (`NotificationNotFoundError` → 404, `NotificationNotDeletableError` → 403) let the route return correct status; the old `if not deleted: 404` couldn't distinguish "missing" from "refused".
- Fails closed: `is_admin` defaults to `False`, and `_role_maps()` swallows lookup errors into empty maps so `is_app_admin` returns `False` on a role-service failure rather than promoting.
- The extracted `is_app_admin` is behaviorally identical to the old inline block in `get_notifications` (`role_by_name.get('Admin')` + `assigned_groups` + `user_info.groups` membership). Since `get_notifications` uses the same notion to *report* `can_delete=True`, what the API reports and what DELETE enforces can no longer drift apart — that's a real correctness argument, not just DRY tidiness.

**One deliberate security decision worth a conscious maintainer sign-off:** this lets admins override `can_delete=False` (the "you must respond" notifications). I agree with the reasoning — `get_notifications` already reports `can_delete=True` to admins, so refusing at DELETE would make the API contradict itself, and there's otherwise no cleanup path for a notification orphaned by a dead workflow. The override is narrow (Admin app-role only), logged at INFO, and audit-recorded as `admin_override`, and the author offered a two-line no-override alternative. Flagging so it's an intentional call, not a side effect.

**Minor / non-blocking:**
- Extra DB round-trip (route does `get_notification_by_id` + access check, then `delete_notification` re-`get`s). Negligible and mirrors mark-as-read.
- The `update_notification` F811 duplicate (L365 & L419) is real — correctly left out of scope and offered as a separate filing.

One note for whoever merges: CI hasn't run (fork PR from a first-time contributor needs workflow approval). My review is a static verification against source plus a read of the 18-case test suite; worth letting them actually execute before merge.

PR #701, opened 2026-08-12
fix(connectors): show BigQuery in Add Connection dropdown

### LINKED ISSUE #683: feat(connectors): dedicated external Unity Catalog connector (OAuth M2M)
## Summary

Add a dedicated connector for **external / remote Unity Catalog workspaces**, configurable from the "Add Connection" dialog and authenticated with **OAuth M2M (service-principal client-credentials)** only.

This is a follow-up to the BigQuery dropdown fix (see PR for `cat-conn-issue`). That PR made fully-implemented connectors selectable again but deliberately did **not** touch the Databricks/UC path, because the current Databricks connector cannot point at any workspace other than the app's own.

## Background: three distinct UC auth modes

We must keep the two **built-in** connections working exactly as today and add a third **external** one:

1. **Built-in (user / OBO)** — uses the requesting user's On-Behalf-Of token (`x-forwarded-access-token`), via `get_obo_workspace_client`. Per-user permissions.
2. **Built-in (app / service principal)** — uses the app's own SP / implicit auth, via `get_workspace_client` (`common/workspace_client.py`). This backs the auto-created system "Databricks UC" connection.
3. **External / remote UC (NEW)** — a *different* workspace, authenticated **only** with provided **OAuth M2M** credentials: `host`, `client_id`, `client_secret`. No PAT (being sunset), no OBO, no ambient app auth.

## Why this doesn't work today

- `DatabricksConnectorConfig` (`src/backend/src/connectors/databricks.py`) has no fields for host or credentials — only an injected `workspace_client`. There's nowhere to store another workspace's connection details.
- `ConnectionsManager._build_connector` (`src/backend/src/controller/connections_manager.py`, ~L153-158) special-cases `connector_type == "databricks"` to **always** reuse the ambient app `self._ws_client` and **discards the stored per-connection config**.
- `get_workspace_client` (`common/workspace_client.py`) is single-workspace: it builds one client from app-level `settings.DATABRICKS_HOST` + PAT / CLI-profile / implicit auth.
- The frontend hides `databricks` from the dropdown entirely: `connection-form-dialog.tsx` (~L247) `filter(ct => ct.connector_type !== 'databricks')`.

## Reference implementation to mirror

The **BigQuery connector** already does per-connection credentialed auth correctly (`src/backend/src/connectors/bigquery.py`, `_ensure_client` L134-175): it resolves credentials per-connection via a priority chain (UC Connection → UC secret scope/key → local file → ADC) and its `is_available` does a real per-config connectivity check. Use it as the template — especially the **UC secret scope/key** pattern for storing the client secret rather than persisting it inline.

## Proposed approach

Prefer a **dedicated connector type** (e.g. `databricks_external`) over overloading `databricks`, to keep the built-in system path (OBO + app SP) untouched and unambiguous. Decide during implementation; a config-driven single type is the alternative.

### Backend

- [ ] New connector class (e.g. `ExternalDatabricksConnector`, `connector_type = "databricks_external"`, `is_selectable = True`) — can subclass/reuse `DatabricksConnector`'s asset logic.
- [ ] Config model with: `workspace_host`, `client_id`, and a **secret reference** for the client secret (`credentials_secret_scope` + `credentials_secret_key`), mirroring BigQuery. Do **not** accept/persist the raw secret inline; add `client_secret`/`credentials` to `_INTERNAL_FIELDS` if ever passed transiently.
- [ ] Build a per-connection `WorkspaceClient(host=..., client_id=..., client_secret=..., auth_type="oauth-m2m")`. Resolve the secret from the UC secret scope via the app's ws client (as BigQuery does with `secrets.get_secret`).
- [ ] `is_available`: real connectivity check against the configured workspace (do NOT gate the dropdown on it — the dropdown now uses `is_selectable`).
- [ ] Register the connector at startup (`utils/startup_tasks.py`, `register_class`).
- [ ] Remove/adjust the `connector_type == "databricks"` short-circuit in `_build_connector` so the new type builds from stored config. Leave the built-in `databricks` (system, OBO/app-SP) path exactly as-is.
- [ ] Config-field hints exposed via `list_connector_types()` for the UI (`_get_config_classes`).

### Frontend

- [ ] Add config-field mapping for the new type in `CONNECTOR_CONFIG_FIELDS` (`connection-form-dialog.tsx`): host, client_id, secret scope, secret key.
- [ ] The existing `!== 'databricks'` filter already hides only the built-in system type; the new `databricks_external` type will appear automatically. Confirm secret fields render as password/secret inputs and are never echoed back.

### Tests

- [ ] Unit: config validation, secret resolution from scope, `is_available` behavior.
- [ ] `_build_connector` builds an OAuth-M2M client for `databricks_external` and still returns the ambient client for system `databricks`.
- [ ] Secret never serialized in `ConnectionResponse` / API output.

## Security notes

- Store the client secret **only** as a UC secret reference; never persist plaintext in the connections table or return it via the API.
- OAuth M2M only — explicitly reject PAT for this connector.

## Acceptance criteria

- A user can create an external UC connection from the dialog by providing host + client_id + secret reference (OAuth M2M).
- The connection lists/reads assets from the remote workspace using those creds.
- The two built-in connections (OBO user, app SP) are unchanged.
- No secret material is ever persisted in plaintext or returned by the API.

larsgeorge-db:
## Summary

The "Add Connection" dropdown was missing the fully-implemented **BigQuery** connector.

**Root cause:** `ConnectionsManager.list_connector_types()` gated the dropdown on each connector's runtime `is_available` property. BigQuery's `is_available` performs a live client check that fails until a project/credentials are configured — which is always the case at startup — so BigQuery was filtered out of the very dialog meant to configure it.

**Fix:** introduce a static `is_selectable` class flag (implementation status, distinct from runtime availability) and gate the dropdown on that instead. Stub/mockup connectors (Snowflake, PowerBI, Kafka) set `is_selectable = False` and stay hidden.

## Changes

- `connectors/base.py` — add `is_selectable: bool = True` on `AssetConnector`, documented vs. runtime `is_available`.
- `connectors/{snowflake,powerbi,kafka}.py` — `is_selectable = False` (stubs stay hidden).
- `controller/connections_manager.py` — filter on `is_selectable` instead of `is_available`; updated docstring.
- `tests/test_connections_manager_connector_types.py` — regression tests (BigQuery listed despite `is_available=False`, default databricks always listed, stubs hidden, metadata populated).

## Notes

- No frontend change needed: BigQuery config fields already exist in `CONNECTOR_CONFIG_FIELDS`, and the existing `!== 'databricks'` filter correctly keeps the system Unity Catalog connection out of the create flow.
- The built-in Databricks/UC connection is intentionally untouched. A dedicated **external Unity Catalog connector (OAuth M2M)** is scoped as a follow-up: #683.

## Testing

`pytest tests/test_connections_manager_connector_types.py` — 4 passed. Verified against the real registry that `bigquery` now surfaces while the three stubs remain hidden.

Closes #683 is **not** intended — that is tracked as a separate follow-up.


mvkonchits-db (approved):
LGTM — the core fix is clean and correct. Swapping the dropdown gate from the runtime is_available check to the static is_selectable class flag is the right depth, and all registered connectors are accounted for. The notes below are low-severity and non-blocking.

1. except Exception: pass can leak a filtered connector into the dropdown — src/backend/src/controller/connections_manager.py:241
If registry.get_connector('snowflake') raises during instantiation, execution jumps to the except before reaching the if not connector.is_selectable: continue guard at line 232. The stub is then appended with a default title and empty capabilities — reappearing in the dropdown despite is_selectable=False, defeating the fix for any stub whose instantiation isn't error-free.

2. Coverage regression on the default-keep branch — src/backend/tests/test_connections_manager_connector_types.py:273
The deleted test_default_connector_is_kept_even_when_unavailable exercised the "keep the default even when it'd otherwise be filtered" branch. The replacement test_default_databricks_always_listed uses a selectable databricks, so that keep-branch is no longer covered. If someone later sets the default connector's is_selectable=False or drops the guard, the primary connection path silently vanishes from the dropdown and no test catches it.

3. Instantiates every connector before checking the static flag — src/backend/src/controller/connections_manager.py:229
is_selectable is a static class attribute, but the loop fully instantiates each connector via get_connector() before checking it, so stubs pay instantiation cost (and any constructor side effects) before being filtered out. Checking the flag on the registered class before instantiating would skip that work and is more in the spirit of why is_selectable was introduced.

4. Duplicated connector-metadata assembly — src/backend/src/controller/connections_manager.py:233
list_connector_types re-builds the display_name/description/capabilities dict that registry.get_connector_info (connectors/registry.py ~319) already assembles. Two hand-rolled copies means a future field addition must be edited in both or the two views drift.

PR #729, opened 2026-08-17
fix(backend): drop stale domain_id kwarg in Admin Team seeding

larsgeorge-db:
## Problem

Startup fails after PR #597 with:

```
TypeError: 'domain_id' is an invalid keyword argument for TeamDb
... StartupSeedError: Default role/team seeding failed
```

PR #597 removed the legacy `domain_id` column from `TeamDb` (domain assignment is now polymorphic via `entity_domain_associations`, `entity_type='team'`), but `SettingsManager.ensure_default_team_and_project` still passed `domain_id=None` to the `TeamDb(...)` constructor. SQLAlchemy's declarative constructor rejects unknown kwargs, so seeding aborts and startup dies.

## Fix

Remove the stale `domain_id=None` kwarg. It was passing `None` anyway, so this is behavior-preserving. `ProjectDb` never had `domain_id`, so it's unaffected.

## Testing

- One-line deletion; startup seeding no longer raises `TypeError`.

### DIFF
--- a/src/backend/src/controller/settings_manager.py
+++ b/src/backend/src/controller/settings_manager.py
@@ -949,7 +949,6 @@ class SettingsManager:
                     name='Admin Team',
                     title='Default Administrator Team',
                     description='Auto-created team for system administrators',
-                    domain_id=None,
                     extra_metadata='{}',
                     created_by='system@startup.ucapp',
                     updated_by='system@startup.ucapp'


mvkonchits-db (approved):
Approved. Verified against source: PR #597 removed the `domain_id` column from `TeamDb`, and this stale `domain_id=None` kwarg was the sole remaining caller that would crash the declarative constructor at startup seeding. Fix is minimal and behavior-preserving (was passing `None` anyway). Confirmed no other `TeamDb(...)` sites are affected — `teams_manager` pops `domain_ids`/`primary_domain_id` before constructing, and `ProjectDb` never had the column.

Non-blocking follow-up: worth adding a unit test around `ensure_default_team_and_project` so this class of column-removal drift is caught in CI.

PR #756, opened 2026-08-25
fix(demo-data): repair packs broken by multi-domain migration; truthful loader + drift-proof teardown

larsgeorge-db:
## Problem

Loading any demo preset reported `"status": "success"` but populated almost nothing in the UI. Root cause: the demo packs (`demo_data_*.sql`) were **stale against the current schema**.

Commit `1688c299` (multi-domain assignment, #520/#597) removed the single-value domain columns — `assets.domain_id`, `data_contracts.domain_id`, `data_products.domain`, `teams.domain_id` — replacing them with the `entity_domain_associations` junction table. The packs still inserted the old columns, so:

- every `assets`/`data_contracts`/`data_products` insert threw `UndefinedColumn`,
- which cascaded FK violations into every dependent child insert (`data_product_teams`, `mdm_*`, `data_contract_*`, `data_profiling_runs`, …).

For retail that was **33 of 66 statements failing silently**. The loader caught each as a warning, counted only successes, and still returned `"success"`. All five vertical presets were broken identically (`migrate` was clean).

## Fixes

**1. Repair the packs (all 5 presets)**
- Strip the removed domain column from the `teams`/`data_contracts`/`data_products`/`assets` inserts.
- Emit `entity_domain_associations` rows (new type code `031`, `is_primary=true`) to preserve each entity's domain link. Products carry a domain *name*, so it's mapped to the domain *id* via the pack's own `data_domains`.
- Refresh the retail type-code legend to reflect post-migration reality (`021`/`025` → `assets`, `022` → `entity_subscriptions`; `023`/`024` retired) and document `031`.

**2. Truthful loader**
- Track failed statements; return `"status": "partial"` with `statements_failed` + a `failures` sample; log at error level. A half-loaded pack no longer masquerades as success.

**3. Drift-proof teardown**
- `clear_demo_data` maintained a hand-written `DELETE ... LIKE '<prefix>%'` list that had itself drifted (never deleted `entity_domain_associations` or `data_contract_quality_checks`). Replaced with a data-driven approach: a new SQL-aware parser (`utils/demo_data_sql.py`) recovers `(table, pk_column, ids)` from each INSERT, and teardown deletes those exact keys in reverse insertion order — FK-safe, and unable to touch built-in rows that merely share an id prefix (so the old `created_by`/`is_system` guards are gone). Same-table deletes are coalesced into one statement so the self-referential `data_domains.parent_id` hierarchy clears under its NO ACTION FK.

The parser handles what a regex can't: `;` inside quoted strings (which had silently truncated the `data_contracts` insert), JSON commas, `COALESCE((SELECT ...))` values, interleaved comments, `ON CONFLICT` tails, and non-`id` PKs (`project_teams`).

## Verification

- Every pack loads under `psql ON_ERROR_STOP=1` with **zero errors** (was 33/66 for retail).
- Row integrity: retail asset rows 11 → 11; total tuples +68, exactly the association rows added.
- Full load + derived-teardown round-trip against Postgres returns the schema to baseline, no FK errors.
- New unit tests `test_demo_data_sql.py` (13 cases incl. all 5 real packs) pass.

Also cleaned up stale "datasets" references (the dropped `datasets`/`dataset_*` tables) in the retail type-code map; no live SQL or app code references the dropped tables.

### DIFF
--- a/src/backend/src/data/demo_data_auto.sql
+++ b/src/backend/src/data/demo_data_auto.sql
@@ -45,11 +45,10 @@ ON CONFLICT (id) DO NOTHING;
 -- 2. TEAMS
 -- ============================================================================
 
-INSERT INTO teams (id, name, title, description, domain_id, extra_metadata, created_by, updated_by, created_at, updated_at) VALUES
-('00100001-0004-4000-8000-000000000001', 'connected-vehicle-platform', 'Connected Vehicle Platform Team', 'Telematics data platform, OTA update orchestration, and remote diagnostics', '00000002-0004-4000-8000-000000000002', '{"slack_channel": "https://company.slack.com/channels/cv-platform", "lead": "cv.architect@oem.com"}', 'system@demo', 'system@demo', NOW(), NOW()),
-('00100002-0004-4000-8000-000000000002', 'adas-engineering', 'ADAS & AD Engineering Team', 'Sensor fusion, perception model training, and autonomous driving validation', '00000003-0004-4000-8000-000000000003', '{"slack_channel": "https://company.slack.com/channels/adas-eng", "tools": ["ROS2", "CARLA", "NVIDIA DRIVE"]}', 'system@demo', 'system@demo', NOW(), NOW()),
-('00100003-0004-4000-8000-000000000003', 'supply-chain-quality', 'Supply Chain Quality Team', 'Supplier PPAP/APQP management, incoming quality, and supply risk analytics', '00000004-0004-4000-8000-000000000004', '{"slack_channel": "https://company.slack.com/channels/scq-team", "responsibilities": ["PPAP", "APQP", "8D", "Supplier Audits"]}', 'system@demo', 'system@demo', NOW(), NOW())
-
+INSERT INTO teams (id, name, title, description, extra_metadata, created_by, updated_by, created_at, updated_at) VALUES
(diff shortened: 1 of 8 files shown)


mvkonchits-db (approved):
Read the full diff — the parser, the teardown refactor, the loader change, and the tests. Solid, well-tested fix.

**Verified:**
- Root cause is real and correctly addressed: the packs still inserted the single-value domain columns dropped by the multi-domain migration (`1688c299`), so ~half of each preset's statements failed while the loader still returned `success`. The fix strips the dead columns and emits `entity_domain_associations` rows (type `031`, `is_primary=true`) to preserve the domain link.
- **Truthful loader** (`partial` + `failures` + error log) is the right call — the silent-success was what masked the drift.
- **Drift-proof teardown**: deriving `(table, pk_column, ids)` from the packs and deleting exact keys in reverse order is both safer than the old prefix `LIKE` list (can't touch built-in rows sharing a prefix) and self-maintaining. The coalesce-per-table rationale for the self-referential `data_domains.parent_id` NO-ACTION FK is sound.
- The parser is genuinely quote/paren/comment-aware (not regex): handles `;` inside quoted strings (the old truncation bug), JSON commas, `COALESCE((SELECT …))`, `''` escapes, `ON CONFLICT` tails, and non-`id` PKs. The 13 unit tests pin exactly these edge cases and parse all 5 real packs. CI (incl. Backend Tests) is green.

One tiny non-blocking nit: `str(err).splitlines()[0]` would `IndexError` on an exception with an empty message — pre-existing pattern in this file and DB errors never carry an empty message, so cosmetic only.

LGTM. Note it's behind `development` by 17 (conflict-free) — leaving the rebase/merge to you as the author.

PR #772, opened 2026-09-04
docs: expand and restructure the documentation site

danielrozo-db:
## Summary

Expands and restructures the Ontos documentation site: a real getting-started section with
both install paths, a consolidated user guide, a full per-resource API reference, and refreshed
introduction/concepts content. Docs-only — no application code changes.

## What changed

**Getting started**
- Add a `getting_started/index.md` landing page covering the two deployment paths and shared
  workspace prerequisites (Foundation Model API, DBSQL endpoint, admin entitlement, UC managed volume).
- Add Marketplace (`install_marketplace.md`) and demo (`install_demo.md`) walkthroughs, including
  loading/deleting bundled demo data; trim `install_databricks.md` to link shared prerequisites.
- Remove the `demo_project.md` / `install_local.md` placeholder stubs; reorder so Marketplace comes first.

**User guide**
- Consolidate the separate Domains, Projects & Teams, Data Contracts, and Data Products pages into
  a single `basic_elements.md` walkthrough of the fundamental elements.
- Add a `discover_data.md` data-discovery guide and refresh pages with screenshots.

**API reference**
- Replace the placeholder `dev_guide/api_reference.md` stub with a `dev_guide/api/` section:
  an `index.md` (interactive OpenAPI surfaces) plus per-resource pages — data contracts, data
  products, data domains, projects, teams, tags, costs — with a `_category_.json`.
- Normalize page titles/headings; point "See also" links at `basic_elements`.
- Remove the `custom_migrations.md` stub.

**Introduction & site**
- Expand `introduction/concepts.md` (adds personas content) and fix typos.
- Landing-page styling tweaks (`src/pages/index.tsx`, `custom.css`, `index.module.css`) and
  `docusaurus.config.ts` updates (exclude list, `deploymentBranch`); minor `faq.md` cleanup.

**Assets**
- Add 35 screenshots referenced by the new walkthroughs.

## Notes

- Docs-only; no changes under `src/`. Rebased cleanly onto `development` (linear history, no merge commits).
- Verified locally with `npm run build` under `onBrokenLinks`/`onBrokenAnchors: 'throw'`.

This pull request and its description were written by Isaac.

PR #789, opened 2026-09-09
ci(docs): deploy Docusaurus site to GitHub Pages

larsgeorge-db:
## What

Adds a GitHub Actions workflow that builds the Docusaurus docs in `website/ontos` and publishes them to **GitHub Pages** at https://databrickslabs.github.io/ontos/.

- `.github/workflows/deploy-docs.yml` — builds with Node 20 and deploys via the Pages artifact (`upload-pages-artifact` → `deploy-pages`). No `gh-pages` branch.
- Triggers on push to `development` (paths under `website/ontos/**` or the workflow itself) plus manual `workflow_dispatch`.
- Removes the stale `deploymentBranch: 'ontos-docs'` from `docusaurus.config.ts` — that setting only applies to the classic `docusaurus deploy` command and is unused by this Actions-based flow (and pointed at a source branch, which would self-overwrite).

The site config (`url`, `baseUrl: /ontos/`, `organizationName`, `projectName`) was already Pages-ready.

## Required repo setting (one-time, manual)

GitHub Pages must be enabled with **Source = "GitHub Actions"** under Settings → Pages. Pages is currently disabled on the repo. Once set, merges to `development` publish automatically.

## Follow-up

Deploy currently triggers from `development` to bootstrap the freshly established docs site (frequent updates expected). **Once the docs reach a publishable steady state, switch the trigger from `development` to `main`** so published docs track `main`.

## Notes

- Local `npm run build` fails on Node 26 (rspack `importFaster` `ERR_MODULE_NOT_FOUND`); CI pins Node 20, so the deployed build is unaffected.

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -1,0 +1,64 @@
+name: Deploy Docs to GitHub Pages
+
+on:
+  push:
+    branches:
+      - development
+    paths:
+      - 'website/ontos/**'
+      - '.github/workflows/deploy-docs.yml'
+  # Allow manual runs from the Actions tab
+  workflow_dispatch:
+
+# Permissions required for the GitHub Pages deployment
+permissions:
+  contents: read
+  pages: write
+  id-token: write
+
+# Allow one concurrent deployment; don't cancel an in-progress one
+concurrency:
+  group: pages
+  cancel-in-progress: false
+
+defaults:
+  run:
+    working-directory: website/ontos
+
+jobs:
+  build:
+    name: Build Docusaurus site
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
+          cache: npm
+          cache-dependency-path: website/ontos/package-lock.json
+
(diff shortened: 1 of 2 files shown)


mvkonchits-db (approved):
Reviewed — correct and low-risk, approving. This is the proper fix for the `deploymentBranch: 'ontos-docs'` line flagged during the #772 review: it reverts that to a commented-out `gh-pages` with an explanatory note and adds a real GitHub Actions Pages workflow.

Verified the two things that could have made this a broken deploy:
- GitHub Pages is already enabled for the repo with `build_type: workflow` (GitHub Actions source) at https://databrickslabs.github.io/ontos/, so `actions/deploy-pages@v4` won't fail for lack of Pages config. Permissions (`pages: write`, `id-token: write`) and the `concurrency: pages` group are right.
- `website/ontos/package-lock.json` exists, so `npm ci` + the npm cache key resolve.

Nice side effect: this puts the Docusaurus `npm run build` (which carries `onBrokenLinks/onBrokenAnchors: 'throw'`) under CI, closing the "docs site isn't gated" gap I noted in #772.

One non-blocking thought: the build runs only post-merge (push to `development`), not on PRs — so a broken doc link fails the deploy *after* merge rather than blocking the PR. If you want pre-merge protection, a `pull_request` build-only job (build without deploy) would catch it earlier. Not required here.

PR #792, opened 2026-09-10
chore(deps): add Dependabot cooldown and cap TypeScript below 7 for docs

### LINKED ISSUE #790: chore(deps): add cooldown to Dependabot config to stop bumps to versions the JFrog registry won't serve yet
## Problem

Dependabot regularly proposes dependency versions **faster than the Databricks JFrog registry (`db-pypi` / `db-npm`) will serve them**, so the bump PRs fail CI on dependency install through no fault of the change itself. Two failures on the same day:

- **[#779](https://github.com/databrickslabs/ontos/pull/779)** — `google-cloud-bigquery` 3.41.0 → **3.45.0**. The version isn't in the JFrog pip mirror yet (only ≤3.44.0 available). `uv`/`pip` resolution fails: `No matching distribution found for google-cloud-bigquery==3.45.0`. Breaks Requirements Lockfile Check, Alembic Single-Head Check, and Backend Tests (Coverage Report fails downstream on the missing artifact).
- **[#773](https://github.com/databrickslabs/ontos/pull/773)** — `lucide-react` 1.16.0 → **1.41.0**. JFrog **Curation** blocks it under a 7-day immaturity cooldown (`{7d-cooldown, Package version is immature 7d, Package version is 5 days old}`): `npm error code E403 ... 403 Forbidden`.

Both are the **same underlying issue**: the registry needs time (mirror sync lag / a 7-day maturity cooldown) before a freshly published version is installable. Any bump to a release younger than ~7 days is liable to red-CI until the registry catches up.

## Proposed fix

Add a **`cooldown`** setting to each `updates` entry in `.github/dependabot.yml` so Dependabot waits ~7 days after a release before opening the PR — matching the JFrog maturity window. This avoids both failure modes above.

```yaml
version: 2
updates:
  - package-ecosystem: "github-actions"
    directory: "/"
    schedule:
      interval: "weekly"
    cooldown:
      default-days: 7

  - package-ecosystem: "pip"
    directory: "/src"
    schedule:
      interval: "weekly"
    cooldown:
      default-days: 7

  # ...same `cooldown` block on the remaining pip (/src/backend, /src/e2e)
  #    and npm (/src/frontend, /website/ontos) entries
```

`cooldown` supports finer knobs if we want them (`semver-major-days`, `semver-minor-days`, `semver-patch-days`, plus `include`/`exclude` package globs). 7 days across the board is the simplest match to the registry's behavior; we can lengthen major-version cooldowns later if desired.

## Acceptance criteria

- [ ] `cooldown` added to all six `updates` entries in `.github/dependabot.yml`
- [ ] `default-days` set to at least 7 (matching the JFrog curation immaturity window)
- [ ] New Dependabot PRs no longer fail install on not-yet-mirrored / immature versions

## Notes

This reduces recurrence but doesn't fully eliminate it — mirror sync can occasionally lag beyond 7 days. Not in scope here, but a longer-term option is getting the JFrog mirror/curation policies aligned so needed versions are available promptly.

larsgeorge-db:
## What

Two changes to `.github/dependabot.yml`:

1. **7-day `cooldown` on all six update entries.** Dependabot keeps proposing versions the Databricks JFrog registry can't serve yet — either the pip mirror hasn't synced the release, or npm curation blocks it under the 7-day immaturity window. This has red-CI'd several dependency PRs (#779 `google-cloud-bigquery`, #777 `mlflow`, #773 `lucide-react`). A 7-day cooldown makes Dependabot wait until the registry can install the version.

2. **Cap `typescript` below 7 for `/website/ontos` only.** TS 7 is the native (Go) compiler rewrite that breaks the JS compiler API Docusaurus 3.x and its tooling depend on; it isn't supported yet (see #747, which bumped to 7.0.2). The docs site has **no CI** to catch the break, so a bad bump merges green silently. The `>=7` cap still lets Dependabot track the supported 6.x line. The main app (`/src/frontend`) is deliberately left free — its required `TypeScript Type Check` job guards it.

## Notes

- The `ignore` is scoped to the `/website/ontos` npm entry, so it does not affect the main app.
- Mitigates but doesn't fully eliminate the registry-lag class of failure; mirror sync can occasionally exceed 7 days.

Closes #790

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,28 +4,48 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    cooldown:
+      default-days: 7
 
   - package-ecosystem: "pip"
     directory: "/src"
     schedule:
       interval: "weekly"
+    cooldown:
+      default-days: 7
 
   - package-ecosystem: "pip"
     directory: "/src/backend"
     schedule:
       interval: "weekly"
+    cooldown:
+      default-days: 7
 
   - package-ecosystem: "pip"
     directory: "/src/e2e"
     schedule:
       interval: "weekly"
+    cooldown:
+      default-days: 7
 
   - package-ecosystem: "npm"
     directory: "/src/frontend"
     schedule:
       interval: "weekly"
+    cooldown:
+      default-days: 7
 
   - package-ecosystem: "npm"
     directory: "/website/ontos"
     schedule:
       interval: "weekly"
+    cooldown:
+      default-days: 7
+    # TypeScript 7.x is the native (Go) compiler rewrite; it breaks the JS
+    # compiler API that Docusaurus 3.x and its tooling rely on, and is not
+    # supported yet. Cap bumps below 7 so Dependabot still tracks the
+    # supported 6.x line but never proposes 7.x. Revisit when Docusaurus v4
+    # (which is expected to carry TS 7 support) lands.
+    ignore:
+      - dependency-name: "typescript"
+        versions: [">=7"]


mvkonchits-db (approved):
Reviewed with the code-review skill — config-only, low-risk. Approving. Three non-blocking notes for your awareness:

1. The TypeScript `>=7` ignore is scoped only to `/website/ontos`; `/src/frontend` (also TypeScript) has no equivalent guard, so a TS7 bump PR can still land there.
2. Dependabot `ignore` conditions also suppress **security** updates in the ignored range — a future advisory fixed only in TS `>=7` would be silently blocked on the docs site (no alerting PR).
3. Worth confirming `cooldown` is supported for every ecosystem listed (incl. `github-actions`); if not, the block is a silent no-op rather than the intended 7-day delay.

None block.

PR #793, opened 2026-09-10
chore(security): allowlist doc-placeholder connection strings in .gitleaksignore

larsgeorge-db:
## Problem

The Databricks pre-push secret scanner (gitleaks) re-flags four example connection strings on **every new-branch push**, forcing contributors to bypass the check with `SKIP_SECRET_SCAN=1`. That's dangerous as a habit — it trains people to disable secret scanning.

Root cause: on a new branch the scanner scans `merge-base(main, branch)..branch`; since branches fork off `development` (which diverged from `main` long ago), that range is huge, so it falls back to scanning all commits authored by the pusher — repeatedly re-hitting old commits that contain example strings.

## Verification

All four flagged strings were inspected and confirmed to be **documentation/example placeholders, not real credentials**:

| File:line | Value | Context |
|---|---|---|
| `.claude/agents/test-engineer.md:822,827` | `postgres:postgres@localhost:5432/test_db` | sample GitHub Actions workflow |
| `docs/SECRETS_SECURITY_SUMMARY.md:62` | `user:password@host:5432/db` | code-comment example |
| `docs/SECRETS_MANAGEMENT_SECURITY_ANALYSIS.md:167` | `user:password@host:5432/db` | self-labeled "Simulated API response" |

## Fix

gitleaks auto-loads `.gitleaksignore` from the repo root, so committing these fingerprints suppresses the false positives for the whole team. Confirmed: this branch pushed cleanly **without** `SKIP_SECRET_SCAN`.

Scoped narrowly to these four fingerprints — any *new* secret still blocks the push.

### DIFF
--- a/.gitleaksignore
+++ b/.gitleaksignore
@@ -1,0 +1,16 @@
+# gitleaks allowlist — false-positive fingerprints.
+#
+# Each entry below is a documentation/example placeholder connection string
+# (not a real credential) that the Databricks pre-push secret scanner
+# (gitleaks) re-flags on every new-branch push, forcing SKIP_SECRET_SCAN=1.
+# Fingerprint format: <commit>:<file>:<rule>:<line>
+#
+# Verified 2026-09-10:
+#   - test-engineer.md: sample CI workflow using postgres:postgres@localhost/test_db
+#   - SECRETS_*.md: user:password@host examples inside secrets-handling docs
+#     (incl. a self-labeled "Simulated API response")
+
+114cb4e9effe26fa7b98e8078007148151f8125d:.claude/agents/test-engineer.md:postgres-connection-string:822
+114cb4e9effe26fa7b98e8078007148151f8125d:.claude/agents/test-engineer.md:postgres-connection-string:827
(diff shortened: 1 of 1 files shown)


mvkonchits-db (approved):
Reviewed with the code-review skill — essentially correct, approving. Verified: both commit SHAs exist and are the commits that introduced the strings, the line numbers match at those commits, and the historical `docs/SECRETS_*.md` paths are right (the later move to `docs/notes/` was a pure rename, invisible to gitleaks' per-commit scan — the obvious trap, avoided). `git log -G` confirms these are the only commits adding these strings, so the allowlist is complete for the rule.

Two low-severity notes, non-blocking:
1. Fingerprints are pinned to exact commit SHAs — any history rewrite (rebase/squash) of those commits silently re-enables the false positives. A gitleaks-config `[allowlist]` path/regex would survive history changes; commit-pinned fingerprints don't.
2. Not verifiable from the repo that the scanner auto-loads a repo-root `.gitleaksignore` and fires exactly the `postgres-connection-string` rule id (no gitleaks config present) — the "pushed cleanly" check is the only evidence it takes effect.

PR #794, opened 2026-09-10
fix(compliance): parse bare True/False as boolean literals in compliance DSL

### LINKED ISSUE #684: [Bug] Compliance DSL parses bare boolean literals (`True`/`False`) as strings, so boolean gates always fail
### In simple terms

Ontos lets an admin write compliance rules like `owner = True` ("this data product must have an owner"). But the rule engine has a parsing bug: when it sees the bare word `True` in a rule, it treats it as the **text string `"True"`**, not the boolean value `True`. The data product's actual "has an owner?" flag is a real boolean, so the engine ends up comparing a real boolean to the word `"True"`, which is never equal — the check **always fails**, for every data product, even ones that clearly do have an owner. Net effect: any compliance gate written with `= True` / `= False` can never pass, so those governance levels show up as **"Not assessed"** for everyone.

### Current Behavior

Compliance rules that compare against a bare boolean literal (e.g. `owner = True`, `has_pii = False`) always evaluate to **False**, regardless of the entity's real value. Any compliance level whose gate uses a boolean literal is never credited, and entities are reported as "Not assessed" for that level.

Confirmed end-to-end in `src/backend/src/common/compliance_dsl.py`:

- **Lexer** (`compliance_dsl.py:238-246`): there is no `TRUE`/`FALSE` keyword and no boolean token type. The bare token `True` is therefore lexed as `Token(IDENTIFIER, "True")`.
- **Parser** (`compliance_dsl.py:415-425`): with no dot/attribute access following it, that identifier becomes `Literal("True")` — a **Python `str`**, not a bool.
- **Evaluator** (`compliance_dsl.py:628-629`): the `EQ` operator does `left == right`, i.e. `True == "True"` → **always `False`** (bool vs str).

Cross-check confirming the entity side is correct: `src/backend/src/controller/entity_dict_builder.py:35` sets `has_owner = bool(product.owner_team_id)` — a genuine `bool`. So the mismatch is purely in the DSL, not in how the entity value is produced.

### Expected Behavior

The DSL should recognise the boolean literals `True` and `False` (ideally `true`/`false` too) as boolean values, so that `owner = True` compares a boolean to a boolean and evaluates correctly. Boolean compliance gates should pass when the entity's corresponding flag is set.

### Steps To Reproduce

1. Define a compliance policy with a rule using a bare boolean literal, e.g. a gate of the form `owner = True` (an "Owner is known" style check).
2. Create/select a data product that genuinely has an owner (`owner_team_id` populated).
3. Run the compliance check for that entity.
4. Observe the gate evaluates to False and the level is reported as "Not assessed", even though the owner is set.

### Root cause summary

`True`/`False` are not tokenised as booleans; they fall through to string literals, so every `= True` / `= False` comparison against a real boolean field is a bool-vs-string comparison that can never be equal.

### Suggested fix direction

Add boolean literals to the lexer/parser (emit a real `bool` for `True`/`False`), and/or make `EQ` coerce/compare booleans consistently. Add unit tests covering `field = True` and `field = False` against both boolean-true and boolean-false entities.

---
*Suggested scope: `scope/compliance`. Applying triage labels is up to maintainers.*

surojitchowdhury:
Fixes #684

## Root cause
The compliance DSL lexer/parser had no boolean literal. In `src/backend/src/common/compliance_dsl.py`:

- The lexer had no `TRUE`/`FALSE` keyword, so a bare `True` was tokenized as `Token(IDENTIFIER, "True")`.
- `Parser.parse_primary` turned a bare identifier (with no following `.`) into `Literal("True")` — a Python **str**, not a bool.
- The EQ evaluator then compared `True == "True"`, which is always `False`.

Net effect: any compliance gate written `field = True` / `field = False` compared a real bool entity property (e.g. `entity_dict_builder.py` sets `has_owner = bool(product.owner_team_id)`) against the string `"True"`/`"False"` and could never pass — levels showed as "Not assessed".

## Fix
- Added `TRUE` and `FALSE` members to the `TokenType` enum.
- Added `'TRUE'`/`'FALSE'` to the lexer `KEYWORDS` set. The lexer already upper-cases identifiers before matching keywords, so `True`/`true`/`False`/`false` all map to the keyword (case-insensitive).
- In `parse_primary`, emit a **real Python bool** — `Literal(True)` for `TRUE` and `Literal(False)` for `FALSE` — placed before the generic IDENTIFIER branch.

EQ/NEQ semantics are unchanged; the fix simply makes the literal a genuine bool so `True == True` / `False == False` work as intended.

## Tests added (`src/backend/tests/test_compliance_dsl.py`, new `TestBooleanLiterals` class)
- `test_lexer_bool_keywords` — `True`/`False`/`true`/`false` lex to `TRUE`/`FALSE` tokens.
- `test_parse_bool_literal_is_python_bool` — parses to a `Literal` holding a real `bool`.
- `test_eq_true_against_bool_field` / `test_eq_false_against_bool_field` — `field = True` / `field = False` evaluate correctly against bool fields.
- `test_neq_true_against_bool_field` / `test_neq_false_against_bool_field` — `!=` behaves correctly.
- `test_lowercase_bool_matches_bool_field` — lowercase `true`/`false` work.
- `test_bool_literal_not_equal_to_string` — regression guard: bool literal is a real bool, not the string `"True"`.

## Docs
Added a "Literal Values" section to `src/docs/compliance-dsl-reference.md` documenting boolean literals (case-insensitive) and noting the `True` vs `'true'` distinction.

## How to verify
```
cd src && hatch -e dev run pytest backend/tests/test_compliance_dsl.py -q
```
All 45 tests pass (37 existing + 8 new).

---

> Re-raised from a branch on this repository (previously #689, opened from a fork) so the required CI workflows can run — fork PRs do not receive the internal JFrog OIDC token, which failed every check at setup before any test ran. Same commits and diff; #689 is closed in favour of this PR.

### DIFF
--- a/src/backend/src/common/compliance_dsl.py
+++ b/src/backend/src/common/compliance_dsl.py
@@ -79,6 +79,8 @@ class TokenType(Enum):
     IDENTIFIER = "IDENTIFIER"
     STRING = "STRING"
     NUMBER = "NUMBER"
+    TRUE = "TRUE"
+    FALSE = "FALSE"
 
     # Punctuation
     LPAREN = "("
@@ -111,7 +113,8 @@ class Lexer:
         'CASE', 'WHEN', 'THEN', 'ELSE', 'END',
         'MATCHES', 'IN', 'CONTAINS', 'AND', 'OR', 'NOT',
         'HAS_TAG', 'TAG', 'LENGTH', 'UPPER', 'LOWER',
-        'PASS', 'FAIL', 'ASSIGN_TAG', 'REMOVE_TAG', 'NOTIFY'
+        'PASS', 'FAIL', 'ASSIGN_TAG', 'REMOVE_TAG', 'NOTIFY',
+        'TRUE', 'FALSE'
     }
 
     def __init__(self, text: str):
@@ -390,6 +393,15 @@ class Parser:
         if token.type == TokenType.NUMBER:
             self.advance()
             return Literal(token.value)
+
+        # Boolean literals (case-insensitive: True/False/true/false)
+        if token.type == TokenType.TRUE:
+            self.advance()
+            return Literal(True)
+
+        if token.type == TokenType.FALSE:
+            self.advance()
+            return Literal(False)
 
         # List literal
         if token.type == TokenType.LBRACKET:
--- a/src/backend/src/common/compliance_dsl.py
+++ b/src/backend/src/common/compliance_dsl.py
@@ -644,6 +644,7 @@ class Evaluator:
                 (type(left) is bool and type(right) in (int, float))
                 or (type(right) is bool and type(left) in (int, float))
             )
+            # Bool-to-bool intentionally falls through because neither type is int or float.
             if has_bool_number_mismatch:
                 return op == TokenType.NEQ
             return left == right if op == TokenType.EQ else left != right
--- a/src/backend/src/common/compliance_dsl.py
+++ b/src/backend/src/common/compliance_dsl.py
@@ -637,10 +637,16 @@ class Evaluator:
         left = self.evaluate(node.left)
         right = self.evaluate(node.right)
 
-        if op == TokenType.EQ:
-            return left == right
-        if op == TokenType.NEQ:
-            return left != right
+        if op in (TokenType.EQ, TokenType.NEQ):
+            # Python considers bool a subclass of int, but DSL booleans must not
+            # compare equal to numeric entity values such as 1 or 0.
+            has_bool_number_mismatch = (
+                (type(left) is bool and type(right) in (int, float))
+                or (type(right) is bool and type(left) in (int, float))
+            )
+            if has_bool_number_mismatch:
+                return op == TokenType.NEQ
+            return left == right if op == TokenType.EQ else left != right
         if op == TokenType.GT:
             return left > right if left is not None and right is not None else False
         if op == TokenType.LT:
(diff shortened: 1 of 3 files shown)


mvkonchits-db (commented):
Reviewed with the code-review skill. The core change is correct and well-tested, but promoting `true`/`false` to reserved boolean literals introduces a **backward-compatibility / semantic regression** worth resolving (or explicitly documenting) before merge — leaving as a comment rather than approving.

1. **String-comparison patterns silently break.** UC tag values and most entity properties are strings, so `ASSERT TAG('certified') = true` previously parsed the RHS as the string `"true"` (string==string, could pass). Now the RHS is bool `True`, so it compares `'true' == True` → always False. Stored rules that used to pass will now silently fail, with no migration/deprecation path.

2. **`bool` is a subclass of `int`, so numeric comparisons coerce.** `ASSERT obj.replica_count = True` now evaluates `1 == True` → True (unintended pass); `obj.n > False` becomes `n > 0`. Previously `False` was the string `'False'`, so these raised/failed cleanly. This can flip a gate from fail to pass.

3. **`true`/`false` become reserved as identifiers too.** Any property name, tag key, or MATCH type spelled `true`/`false` (any case) now lexes as `TRUE`/`FALSE` and fails to parse (`Expected IDENTIFIER`), where it used to be a valid identifier.

4. Docs note the quoted-vs-unquoted distinction but don't warn that `TAG()`/string-property comparisons must now use quoted `'true'`/`'false'` — the exact pitfall this introduces.

Suggestion: keep the fix, but either restrict bool coercion to genuinely boolean entity fields, or clearly document that string/tag comparisons must be quoted, and sanity-check existing stored rules for the `= true`/`= false` pattern. Not asserting these all fire in practice — flagging for you to confirm against real rules.

larsgeorge-db (approved):

PR #796, opened 2026-09-10
fix(access): cap approve/reject dialogs to viewport with internal scroll

### LINKED ISSUE #687: [Bug] Access approve/reject dialogs lack max-height/scroll — Approve/Deny buttons fall off-screen on tall content
### In simple terms

When a Data Owner opens the **Approve / Reject** dialog for an access request (from the notification bell), the dialog box has no height limit and doesn't scroll inside. If the request has a lot of detail, the box grows taller than the screen, its top and bottom get cut off, and the **Approve / Deny buttons fall off the bottom of the screen** where you can't click them. The only workaround today is to zoom the whole browser out until the buttons come back into view. The buttons should always be reachable — the dialog should cap its height and scroll its content, keeping the action buttons visible.

### Current Behavior

Both notification-launched approve/reject dialogs render inside a fixed, vertically-centered `DialogContent` with **no `max-height` and no internal `overflow` scroll region**, and their action buttons are direct layout children (not a pinned/sticky footer). With tall content the dialog grows past the viewport, its top and bottom are clipped, and the Approve/Deny buttons become unreachable — the user must zoom the browser out to click them.

Affected components (both launched from the notification bell at `src/frontend/src/components/ui/notification-bell.tsx:493` and `:512`):

- `src/frontend/src/components/access/handle-access-grant-dialog.tsx` — legacy access-grant path; buttons **Deny / Approve** (`:305`, `:314`).
- `src/frontend/src/components/workflows/workflow-approval-response-dialog.tsx` — workflow-approval path; buttons **Reject / Approve** (`:411`, `:424`).

Layout evidence — base `DialogContent` at `src/frontend/src/components/ui/dialog.tsx:38-52`:
```
"fixed left-[50%] top-[50%] z-50 grid w-full max-w-lg translate-x-[-50%] translate-y-[-50%] gap-4 border bg-background p-6 shadow-lg ... sm:rounded-lg"
```
It is `fixed` + `translate-y-[-50%]` (centered) with **no `max-h`** and **no `overflow`** — the panel's height is purely its content's natural height. Neither dialog adds bounding:
- `handle-access-grant-dialog.tsx:160` → `<DialogContent className="sm:max-w-[560px]">` (width only).
- `workflow-approval-response-dialog.tsx:322` → `<DialogContent className="sm:max-w-md">` (width only).

The footer buttons are plain grid children that grow with the content, not a pinned footer.

### Expected Behavior

The dialog should cap its height to the viewport (e.g. `max-h-[85vh]`) and make its body scroll (`overflow-y-auto`), with the Approve/Deny (Reject) actions pinned in a footer that stays visible. Action buttons must always be reachable without zooming the browser out.

### Steps To Reproduce

1. As a Data Owner, receive an access request with a lot of detail (long justification / many fields).
2. Open the Approve/Reject dialog from the notification bell.
3. Observe the dialog grows taller than the viewport; its content does not scroll internally and the Approve/Deny buttons are pushed off-screen.
4. Note the only workaround is zooming the browser out to reach the buttons.

### Root cause summary

The shared `DialogContent` and both approval dialogs set width only — no `max-height`, no internal `overflow-y` scroll, and no pinned footer — so tall content overflows the viewport and clips the action buttons.

### Suggested fix direction

Add a viewport-bounded height + internal scroll to the dialog body and pin the action row as a sticky/non-scrolling footer. Since the base `DialogContent` is shared, decide whether to fix it centrally (affects all dialogs) or per-dialog for these two. Verify with tall content that buttons stay visible at 100% zoom.

---
*Suggested scope: `scope/entitlement` / access. Applying triage labels is up to maintainers.*

surojitchowdhury:
Fixes #687

## Summary

The notification-launched Approve/Reject access dialogs had no height cap and no internal scroll, so with tall content the dialog grew past the viewport and the Approve/Deny (Reject) buttons fell off-screen (only reachable by zooming the browser out). This caps each dialog to the viewport, scrolls the body internally, and pins the header + action footer so the buttons stay reachable at 100% zoom for any content height.

## Before / After (UI)

Same access-request dialog, same content, at a short viewport (~620px tall).

**Before (`development`)** — the dialog overflows the viewport: the header is clipped at the top and the **Deny / Approve buttons fall off-screen** with no way to scroll to them (workaround was zooming the browser out).

Before: approve/deny buttons off-screen

**After (this PR)** — the dialog caps to the viewport: the header + action footer are pinned and the body scrolls internally, so **Deny / Approve stay reachable** at 100% zoom.

After: buttons pinned, body scrolls

## Change (targeted per-dialog — shared `ui/dialog.tsx` untouched)

Both `handle-access-grant-dialog.tsx` and `workflow-approval-response-dialog.tsx`:
- `DialogContent`: add `flex flex-col max-h-[85vh] overflow-hidden` (flex column capped to 85% of viewport height; `flex` overrides the base `grid` via tailwind-merge, base `gap-4` still applies).
- Header: `shrink-0` (stays pinned).
- Body: `flex-1 overflow-y-auto` so only the middle scrolls; `px-1 -mx-1` for focus-ring breathing room. In the workflow dialog a scroll wrapper was added around the content so the footer remains a non-scrolling sibling.
- `DialogFooter`: `shrink-0` (pinned, always visible).

### Scope decision
Fixed per-dialog rather than in the shared `DialogContent`, whose base is used by every dialog in the app — a central max-h/overflow/flex change would have broad blast radius. The two affected dialogs already have header/body/footer structure, so a ~10-line scoped change fully resolves #687 without risking other dialogs.

## Gates
- `yarn type-check` (tsc --noEmit): pass.
- `yarn build`: success (pre-existing chunk-size warning only).
- eslint: no new problems (the two files' existing warnings are pre-existing, identical on untouched HEAD).

---

> Re-raised from a branch on this repository (previously #726, opened from a fork) so the required CI workflows can run — fork PRs do not receive the internal JFrog OIDC token, which failed every check at setup before any test ran. Same commits and diff; #726 is closed in favour of this PR.

### DIFF
--- a/src/frontend/src/components/access/handle-access-grant-dialog.tsx
+++ b/src/frontend/src/components/access/handle-access-grant-dialog.tsx
@@ -157,8 +157,8 @@ export default function HandleAccessGrantDialog({
 
   return (
     <Dialog open={isOpen} onOpenChange={onOpenChange}>
-      <DialogContent className="sm:max-w-[560px]">
-        <DialogHeader>
+      <DialogContent className="sm:max-w-[560px] flex flex-col max-h-[85vh] overflow-hidden">
+        <DialogHeader className="shrink-0">
           <DialogTitle className="flex items-center gap-2">
             <Shield className="h-5 w-5 text-primary" />
             Handle Access Request
@@ -168,7 +168,7 @@ export default function HandleAccessGrantDialog({
           </DialogDescription>
         </DialogHeader>
 
-        <div className="space-y-4">
+        <div className="space-y-4 flex-1 overflow-y-auto px-1 -mx-1">
           {/* Request Details */}
           <div className="p-4 bg-muted/50 rounded-lg border space-y-3">
             <div className="flex items-center gap-2">
@@ -287,7 +287,7 @@ export default function HandleAccessGrantDialog({
           </div>
         </div>
 
-        <DialogFooter className="gap-2 sm:gap-2 flex-col sm:flex-row">
+        <DialogFooter className="gap-2 sm:gap-2 flex-col sm:flex-row shrink-0">
           <Button
             variant="outline"
             onClick={() => onOpenChange(false)}
--- a/src/frontend/src/components/workflows/workflow-approval-response-dialog.tsx
+++ b/src/frontend/src/components/workflows/workflow-approval-response-dialog.tsx
@@ -319,17 +319,18 @@ export default function WorkflowApprovalResponseDialog({
 
   return (
     <Dialog open={isOpen} onOpenChange={onOpenChange}>
-      <DialogContent className="sm:max-w-md">
-        <DialogHeader>
+      <DialogContent className="sm:max-w-md flex flex-col max-h-[85vh] overflow-hidden">
+        <DialogHeader className="shrink-0">
           <DialogTitle>{title}</DialogTitle>
           <DialogDescription>{description}</DialogDescription>
         </DialogHeader>
         {loadingConfig ? (
-          <div className="flex items-center justify-center py-8">
+          <div className="flex flex-1 items-center justify-center py-8">
             <Loader2 className="h-8 w-8 animate-spin text-muted-foreground" />
           </div>
         ) : (
           <>
+            <div className="flex-1 overflow-y-auto space-y-4 px-1 -mx-1">
             {detailRows.length > 0 && (
               <div className="rounded-md border bg-muted/30 p-3 space-y-1.5">
                 <div className="text-xs font-medium text-muted-foreground uppercase tracking-wide">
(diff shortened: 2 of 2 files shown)


mvkonchits-db (approved):
Reviewed with the code-review skill — no correctness bugs, approving. Verified the parts most likely to break: the `grid`→`flex flex-col` swap is handled by tailwind-merge (drops `grid`, keeps `gap-4`); the classic flexbox scroll gotcha is covered because the scroll child has `overflow-y-auto` (auto min-size 0); and inner Radix `Select` dropdowns are portaled, so `overflow-hidden` won't clip them.

Cleanup/altitude notes, non-blocking:
1. The fix is applied per-dialog, but the unbounded height is a property of the shared shadcn `DialogContent` (~76 of 128 `DialogContent` usages still lack a viewport cap). Adding the cap + scroll behavior once in `ui/dialog.tsx` would fix all dialogs and avoid repeating this.
2. The scroll-wrapper recipe (`flex-col max-h/overflow-hidden` + `flex-1 overflow-y-auto` + `shrink-0` header/footer) is duplicated verbatim across both files — a small shared wrapper would prevent drift.
3. The new wrapping `<div>` left its ~65 enclosed lines at the old indentation; a formatter pass would make the nesting readable.

PR #815, opened 2026-09-16
fix(docs): build Docusaurus without faster/strict-MDX v4 flags

larsgeorge-db:
## What

Third and final fix for the Pages deploy. With the action pins resolved, the build reached Docusaurus and failed:

> To enable Docusaurus Faster options, your site must add the `@docusaurus/faster` package as a dependency. — `Cannot find package '@docusaurus/faster'`

**Root cause:** `future.v4: true` is a shorthand that enables *all* v4 future flags, two of which this site can't satisfy:

1. **`fasterByDefault`** turns on the Docusaurus Faster (rspack) bundler, which requires the `@docusaurus/faster` package. That package isn't installable in our environment — its transitive dep `@module-federation/runtime-tools@0.22.0` isn't on the Databricks npm proxy, and public npm is firewalled, so a valid lockfile can't be produced.
2. **`mdx1CompatDisabledByDefault`** enables strict MDX. That surfaced a *latent* error the rspack crash had been masking: the `{#dbsetup_stepN}` heading-id anchors in `getting_started/install_databricks.md` are parsed as JS expressions under strict MDX and fail acorn parsing.

## Fix

Enable the v4 future flags individually, with those two set to `false`:

```ts
future: {
  v4: {
    removeLegacyPostBuildHeadAttribute: true,
    useCssCascadeLayers: true,
    siteStorageNamespacing: true,
    fasterByDefault: false,          // keep webpack bundler
    mdx1CompatDisabledByDefault: false, // keep MDX v1 heading-id anchors
  },
},
```

The site stays otherwise v4-ready. **Verified with a clean local `npm run build`** (server + client compiled, static files generated).

## Follow-ups (not blocking)

- To adopt Docusaurus Faster later, add `@docusaurus/faster` once the proxy mirrors its deps (or build with public-npm access), then flip `fasterByDefault` back on.
- To adopt strict MDX, migrate the `{#id}` heading anchors in `install_databricks.md` (Docusaurus supports them, but strict MDX needs an alternative), then flip `mdx1CompatDisabledByDefault` on.


mvkonchits-db (approved):
✅ **Approved**

Excellent documentation and thoughtful configuration. This PR correctly migrates from blanket `v4: true` to explicit flag management.

**Verification Summary:**
- All 5 Docusaurus v4 future flags accounted for
- TypeScript config syntax valid
- Rationale clear: disables `fasterByDefault` (rspack dep unavailable on npm proxy) and `mdx1CompatDisabledByDefault` (breaks heading anchors in install docs)
- Three v4-ready flags enabled; two pragmatically disabled pending environment fixes
- CI passing (TypeScript check, frontend build, all tests)

**Root Cause Fix:** Selective flag disabling properly addresses the build failure instead of blanket enable/disable.

Minimal, focused, ready to merge.

PR #816, opened 2026-09-17
chore(deps): ignore hatchling >=1.32 in Dependabot (breaks CI)

larsgeorge-db:
hatchling 1.32 changed the readme-path validation and breaks our CI build (see #755, now closed). This adds a Dependabot `ignore` for hatchling `>=1.32` in the `/src` pip block so it stops being re-proposed on the supported line.

Revisit once we ship a project-local readme and can uncap.


mvkonchits-db (approved):
✅ Approved — all CI passing.

PR #818, opened 2026-09-17
ci(docs): run Pages deploy on protected runner group

larsgeorge-db:
## What

The docs build now passes, but the **deploy** step fails ([run](https://github.com/databrickslabs/ontos/actions/runs/35211492849)):

> HttpError: ... the `databrickslabs` organization has an IP allow list enabled, and your IP address is not permitted to access this resource.
> Error: Failed to create deployment (status: 403) ... Ensure GITHUB_TOKEN has permission "pages: write".

**Root cause:** stock GitHub-hosted `ubuntu-latest` runners have IPs that aren't on the `databrickslabs` org IP allow list, so `actions/deploy-pages`' Pages API call is rejected. (The `403 pages: write` line is a secondary symptom of the same block, not a permissions-config problem.)

This is a known constraint in this org — `backmerge-main-to-development.yml` documents it verbatim and `test-coverage.yml` already runs on the protected runner group. I copied the sibling repo `databrickslabs/dqx`, which builds its Docusaurus docs on `databrickslabs-protected-runner-group` for the same reason.

## Fix

Move both jobs from `ubuntu-latest` to:

```yaml
runs-on:
  group: databrickslabs-protected-runner-group
  labels: linux-ubuntu-latest
```

These are IT-managed runners with allow-listed IPs, so the Pages API deployment succeeds. No other changes; this is the last blocker to publishing at https://databrickslabs.github.io/ontos/.


mvkonchits-db (approved):

mvkonchits-db (approved):

mvkonchits-db (approved):
✅ Approved — all CI passing.

PR #819, opened 2026-09-17
fix(data-contracts): make uploaded/imported ODCS drafts visible to their creator

larsgeorge-db:
## Problem

Uploading an ODCS contract from YAML (or pasting ODCS JSON) reported success, but the contract never appeared in the Data Contracts list for the user who uploaded it.

## Root causes

Two independent causes, both fixed here:

1. **`create_from_upload()` never set `draft_owner_id`.** The uploaded contract is `draft` with no `owner`, so the owner resolves to the uploader's username, which `_resolve_team_name_to_id()` can't match to a team → `owner_team_id = None`. Under the PRD #442 role-aware visibility filter, a `draft` row is only visible to a non-admin when the version family is *elevated* — i.e. `draft_owner_id == caller` or `owner_team_id ∈ caller's teams`. With both `None`, the draft is filtered out of the list for its own uploader. The regular `create()` path already stamps `draft_owner_id = current_user` for exactly this reason; `create_from_upload()` (and the `odcs/import` route that delegates to it) never got the same treatment.

2. **The list route used the wrong admin check.** `GET /data-contracts` resolved admin via `is_user_admin` (workspace-group membership only), whereas `GET /data-products` uses the role-aware `is_user_feature_admin`. A user whose admin rights come from an Ontos role — not the workspace `admins` group — was therefore treated as a consumer on Contracts and had their own draft/proposed contracts hidden, even though the same identity sees everything on Products. This is the asymmetry behind "the product shows up but the contract doesn't".

Either fix alone surfaces the contract; together they cover both non-admin owners and role-based admins.

## Changes

- `data_contracts_manager.py` — `create_from_upload()` stamps `draft_owner_id = current_user` for ownerless drafts (covers both the upload and paste-JSON import paths).
- `data_contracts_routes.py` — `get_contracts` switched to `await is_user_feature_admin(..., feature_id='data-contracts')`, matching the data-products list route (added `request: Request` to the signature).

## Verification

- End-to-end run of the real `create_from_upload()` against the reported YAML: now yields `draft_owner_id = <uploader>` and the contract passes the visibility filter for the non-admin uploader.
- Unit tests: `test_version_visibility.py` + `test_is_user_feature_admin.py` — 21 passed.
- The `test_data_contracts_routes.py` integration failures are pre-existing (identical 503/500 on `development` without these changes, a test-harness issue) and unrelated to this PR.


mvkonchits-db (approved):

mvkonchits-db (approved):

mvkonchits-db (approved):
✅ Approved — all CI passing.

PR #820, opened 2026-09-18
fix(data-products): render customProperties/authoritativeDefinitions and preserve null values

larsgeorge-db:
## Problem

Uploading an ODPS data product from YAML succeeded, but the detail view "lost" fields — most visibly **customProperties**. Diagnosis showed the backend persisted and returned the data correctly; the detail view simply had no panel for it (nor for `authoritativeDefinitions`). Two related backend issues surfaced alongside.

## Changes

**Frontend — `data-product-details.tsx`**
- New read-only **Custom Properties** and **Authoritative Definitions** panels, appended after Support Channels. Gated to the **Large** view only, mirroring the contract detail view's `shouldShowSection` behavior. Each renders only when non-empty.
- Null values (and the legacy `"null"` string, see below) render as a muted italic *null* rather than clutter.

**Frontend — `data-products.tsx`**
- Upload success toast read `response.data.count`, but the endpoint returns a `List[DataProduct]`; it always showed 0. Now reads the list length.

**Backend**
- `data_products_repository.py`: null custom-property values were stored as the literal string `"null"` (`json.dumps(None) == "null"`). New `_serialize_custom_property_value` helper (used by create and update) stores them as SQL NULL.
- `models/data_products.py`: `CustomProperty.value` is now optional, with a before-validator that maps the legacy `"null"` string back to `None` on read — so existing rows also present real nulls.
- `data_product_routes.py`: `POST /data-products/upload` now sets `response_model_by_alias=False`, so its response emits camelCase field names (`customProperties`, `outputPorts`, …) consistent with the GET routes and the frontend TS types (it previously emitted the snake_case validation aliases).

## Verification

- **Browser (Playwright):** uploaded the reported product YAML through the UI, opened the detail view, switched to Large — the **Custom Properties (13)** panel renders with real values in mono and all 9 unresolved props as muted *null*; placed after Support Channels, before Subscribers. (Authoritative Definitions panel correctly hidden — the test file has none.)
- **Backend NULL round-trip:** null props now store `None` and read back `None`; non-null values preserved.
- Frontend `tsc --noEmit`: 0 errors. Repository/schema unit tests: 30 passed. (`test_data_products_manager.py` has 6 failures that are **pre-existing** — identical on `development` — unrelated to this change.)

## Notes / follow-ups (not in this PR)

- The two new panels are **Large-view only** to match the contract detail view. Since a freshly uploaded product has no owner, its uploader defaults to Medium/Minimal and must click **L** to see them — a deliberate, discussed choice for cross-entity consistency.
- Sibling product routes (`POST /data-products`, `PUT /{id}`, `/from-contract`, `/published`, `/my-subscriptions`, `/versions`, `/clone-for-editing`) still serialize `by_alias=True` — the same latent casing inconsistency, worth a follow-up sweep.


mvkonchits-db (approved):

mvkonchits-db (approved):

mvkonchits-db (approved):
✅ Approved — all CI passing.

PR #822, opened 2026-09-18
fix(frontend): fix Apps build ERESOLVE via legacy-peer-deps

larsgeorge-db:
## Problem

The Databricks Apps deployment fails at the frontend build step (`scripts/build_static.sh` → `npm ci` / `npm install`) with:

```
npm error code ERESOLVE
While resolving: @hookform/resolvers@5.9.1
  Found: ajv-formats@3.0.1  (root project depends on ajv-formats@^3.0.1)
  Could not resolve:
  peerOptional ajv-formats@"^2.1.1" from @hookform/resolvers@5.9.1
```

`@hookform/resolvers@5` declares an **optional** peer dependency on `ajv-formats@^2`, while the app depends on `ajv-formats@^3` directly. npm 7+ strict peer resolution refuses this on a fresh `npm install`, so the Apps build aborts before Vite ever runs.

It passed locally only because `npm run build` reused an existing `node_modules` and `npm ci` installs from the lockfile without re-resolving peers — the conflict surfaces only on the fresh install the build container performs.

## Why this fix

The optional ajv peer is **unused**: forms use `zod` with `react-hook-form`, and `ajv-formats@3` is a direct dependency for the app's own JSON-schema validation. So the correct resolution is to accept the optional-peer mismatch rather than downgrade `ajv-formats`.

Adding `src/frontend/.npmrc` with `legacy-peer-deps=true` applies to **every** npm invocation — the container's `npm ci`, the `npm install` fallback, and local installs — keeping local and CI behavior consistent.

## Change

- Add `src/frontend/.npmrc` (`legacy-peer-deps=true`) with an explanatory comment.

Verified locally: plain `npm install --dry-run` reproduces the ERESOLVE; `--legacy-peer-deps` resolves cleanly. Confirmed fixed in the actual Apps deployment build.


mvkonchits-db (approved):
✅ Approved — all CI passing.

Issue #763, opened 2026-08-27
[PRD]: Large-account scale, async importers, and UC domain sync


larsgeorge-db:
PRD: `docs/prds/prd-large-account-scale-and-uc-sync.md`.

## Sub-tasks
- [ ] #757 Make Schema Importer child limit configurable (item 1)
- [ ] #758 Async Schema Importer with size detection + background run (item 2)
- [ ] #759 Document Ontos MCP with Genie (item 3)
- [ ] #760 Approving a role access request does not grant the role (item 4)
- [ ] #761 Domain sync with Unity Catalog — import + export (item 5)
- [ ] #762 Sort Schema/Domain Importer entries alphabetically (item 6)

## Problem Statement
A large-enterprise customer running Ontos against a big Unity Catalog estate hit a cluster of scale limitations: the Schema Importer silently caps at 500 children, imports run synchronously and block the UI, there is no guide for using the Ontos MCP with Genie, role-access approvals don't actually grant access, domains can't be synced with UC, and importer lists aren't sorted. See PRD `docs/prds/prd-large-account-scale-and-uc-sync.md`.

## Proposed Solution
Ship the six sliced items above. Cross-cutting decisions (dual sync/async execution model reusing `JobsManager` + notifications, generic `app_settings` for new numeric settings, `assigned_users` dependency for #760) are recorded in the PRD.

Suggested build order: **#757 first** (settings plumbing + browse-response shape shared with #762) → **#758** (async model reused by #761) → **#761** → **#759 / #760 / #762** independent. **#760 depends on `assigned_users` from `prd-individual-user-role-assignment.md`.**

## Additional Context
Reuses existing infrastructure: `JobsManager` (background thread + Databricks Job + `workflow_job_runs`), notification bell (`JOB_PROGRESS`, 60s polling), `DomainExportAdapter` + `uc_tag_sync.reconcile_tags`, generic `app_settings` key-value store. All six issues originate from an external customer request.

PR #823, opened 2026-09-19
feat(schema-importer): configurable per-path child fetch limit (#757)

### LINKED ISSUE #757: Make Schema Importer child limit configurable

## Problem statement
The Schema Importer fetches at most 500 assets per path via a hardcoded `ListAssetsOptions(limit=500)` at two sites in `src/backend/src/controller/schema_import_manager.py` (browse ~L163, collection ~L698). Schemas/catalogs with >500 children silently lose the overflow — never fetched, no "load more", no truncation indication. On large estates this hides real assets.

## Proposed Solution
- Add General Settings key `schema_import_child_limit` in the generic `app_settings` key-value table (no migration). Config field + env default (500), loaded on startup, exposed via `GET /api/settings`, validated on `PUT /api/settings` to `1..10000` (the existing `ListAssetsOptions` bound in `connectors/base.py`).
- `SchemaImportManager` reads the configured value at both fetch sites instead of `500`.
- Add a truncation signal to the browse response (e.g. `truncated`/`truncated_at`) so the UI can show "showing first N — raise the limit in Settings".
- Frontend: add the numeric field to General Settings (`components/settings/general-settings.tsx` + `app-settings-store`) and surface the truncation notice in the browse tree.
- Default 500 → no behavior change until an admin opts in.

## Additional Context
No pagination — raising the cap is the mechanism; continuation past 10000 is an explicit follow-up. Shares the browse-response change with the alphabetical-sorting issue; land together or this one first. Tests: manager uses configured limit at both sites; settings validation rejects out-of-range; browse reports truncation at the cap.

larsgeorge-db:
## What & why

The Schema Importer hardcoded a **500-object cap per path** (`ListAssetsOptions(limit=500)`) at both the browse and import-collection fetch sites. On large Unity Catalog estates, schemas or catalogs with more than 500 children silently lost the overflow — the extra objects were never fetched, there was no "load more", and nothing told the operator the list was truncated.

This makes the cap a configurable General Setting and makes truncation visible.

Part of the large-account scale batch (epic #763). PRD: `docs/prds/prd-large-account-scale-and-uc-sync.md`.

## Changes

- **Config + setting**: new `SCHEMA_IMPORT_CHILD_LIMIT` (default 500), persisted in `app_settings`, validated to the connector contract's `1..10000` range. `SchemaImportManager` reads it per request at both fetch sites; a stale/invalid persisted value is clamped so it can never violate `ListAssetsOptions`.
- **Truncation signal**: `truncated` / `truncated_at` added to `BrowseResponse`. Set when the connector returns a full page.
- **Frontend**: numeric field in **Settings → General → Schema Importer**, wired through fetch/save; truncation notices shown at the browse root and under each expanded node when its children were capped.
- **Tests**: limit resolution (default / override / clamp / invalid fallback), truncation signaling on/off, and settings validation (rejects out-of-range and non-integers; persists valid values).

## Notes

- **No pagination** — raising the cap is the mechanism; the truncation signal makes the `1..10000` ceiling honest. Continuation past 10000 is a documented follow-up.
- Default remains 500, so behavior is unchanged until an admin opts in.
- The umbrella PRD + issues draft for the batch land in this PR under `docs/prds/`.
- Pre-existing unrelated failure `test_get_settings_returns_dict` (mock fixture missing `WORKSPACE_DEPLOYMENT_PATH`) reproduces on `development` and is untouched here.

Closes #757


mvkonchits-db (approved):

mvkonchits-db (approved):

mvkonchits-db (approved):
✅ Approved — all CI passing.

Issue #827, opened 2026-09-19
Optional Databricks Job execution backend for the async Schema Importer


larsgeorge-db:
## Problem statement

The async Schema Importer (#758) runs imports in an in-app daemon thread — the pattern the codebase already uses for long-running UC work (Ontology Generator). This is sufficient for the reported large-account cases, but an in-app thread does not survive an app restart/redeploy and shares the app's process resources.

## Proposed Solution

Add an optional **Databricks Job** execution backend for very large imports, selectable alongside the in-app thread:

- Package the import as an installed workflow (reuse `JobsManager` install/run/poll + `workflow_job_runs` tracking).
- Pass the `ImportRequest` as job parameters; the job writes progress/results back (DB or callback).
- Surface it through the same `schema_import_runs` run-state + notification UX (#758), so the frontend is agnostic to the backend.

## Additional Context

Deferred from #758 (agreed: "in-app now, Databricks Job as follow-up"). Only worth building if a real need appears — imports that must survive app restarts, or offloading very large runs off the app process. Part of epic #763 context.

Issue #826, opened 2026-09-19
All async/background processes should report completion via the notification system


larsgeorge-db:
## Problem statement

Ontos runs several operations as async/background work (in-app daemon threads today): the Ontology Generator, Term Mapping, and now the async Schema Importer (#758). Their completion-reporting is inconsistent — some surface progress only via status-polling endpoints, so a user who navigates away never learns the job finished.

## Proposed Solution

Establish a convention (and retrofit existing async features) that **every** async/background process reports completion through the in-app notification system:

- On start: create a `JOB_PROGRESS` notification addressed to the initiating user.
- On progress (optional): update the notification's `data.progress`/`status`.
- On terminal state: update it to a `SUCCESS` / `ERROR` (or completed/failed) message with a deep link to the result.

The async Schema Importer (#758) already implements this pattern (see `SchemaImportRunsManager._create_progress_notification` / `_finalize_notification`) and can serve as the reference. Candidates to retrofit: Ontology Generator (`OntologyGeneratorManager`), Term Mapping, and any future long-running job.

Consider extracting a small shared helper (create/update/finalize a run-progress notification) so each async manager doesn't re-implement it.

## Additional Context

Raised during #758 review. Part of the large-account scale batch (epic #763) context, but broader — applies to all async features.

PR #817, opened 2026-09-17
feat(mcp): keyless read-only MCP access via a default token

larsgeorge-db:
## What & why

Genie's native MCP connector carries a single credential in the `Authorization` header, so it clears the Databricks app-gate but is then rejected by Ontos's inner `X-API-Key` check (`-32001`). This resolves an app-gate-authenticated MCP request that carries **no** `X-API-Key` to an admin-designated **keyless default** token, stamped with the caller's forwarded email for audit. The default token's scopes bound exactly what the keyless caller may do; keyless access is off until such a token exists.

Implements PRD #730.

## Approach

- The MCP token is a tool-authz + identity + lifecycle gate (not a data-access gate). `_has_scope` is pure, so a resolved principal's scopes flow through the existing tool gating unchanged.
- New `resolve_mcp_principal(db, x_api_key, request)` replaces `validate_api_key` at the GET/POST/DELETE call sites: key present → validate as before; key absent + forwarded identity → resolve the single active `is_keyless_default` token via `MCPTokensManager.resolve_keyless_default(email)`. Header-trust only (no OBO/SDK round-trip), matching the regular HTTP path's trust in the same forwarded headers.
- **Switch = presence:** keyless works iff an active, non-expired default token exists — no separate settings flag. Coarse revocation = deactivate the row; attribution preserved via `created_by = X-Forwarded-Email`.

## Changes

**Backend**
- `mcp_tokens`: `is_keyless_default` column + migration `m1_mcp_keyless_default` (descends from `l1_entity_domain_associations`; single head).
- Repo `get_keyless_default` / `set_keyless_default` / `clear_keyless_default` — single-active-default enforced in code.
- Manager `resolve_keyless_default`, `set_keyless_default`, `clear_keyless_default`; `generate_token(is_keyless_default=…)`.
- Admin routes: `is_keyless_default` on create + list/get; `POST /api/mcp-tokens/{id}/keyless-default`, `DELETE /api/mcp-tokens/keyless-default` (admin-only, audited).
- Fix `MCPTokenDb.is_expired` to treat a naive `expires_at` as UTC.

**Frontend**
- `is_keyless_default` on the token type; keyless-default badge; star/unstar row actions; create-dialog checkbox with a warning that `global_search` bypasses per-user ACLs (recommend read-only scopes). en i18n strings.

## Security notes

[...]

PR #828, opened 2026-09-19
feat(schema-importer): async background import with progress + notifications (#758)

### LINKED ISSUE #758: Async Schema Importer with size detection and background run

## Problem statement
`execute_import` (and preview collection) run inside the HTTP request, fetching per-asset metadata while creating each Ontos asset. On a large estate this takes minutes and blocks the UI with no progress, no background option, and no completion signal.

## Proposed Solution
- Detect estate size up front (reuse the preview/collection count). Below a configurable threshold, keep the current synchronous path unchanged.
- At/above threshold, return immediately with a run id and execute in the background, **user-selectable** between an **in-app background thread** (medium) and a **Databricks Job** (very large), reusing the `JobsManager` patterns (`_monitor_job_progress`, install/run/poll, `workflow_job_runs`).
- Progress via `NotificationType.JOB_PROGRESS`; terminal `SUCCESS`/`ERROR` notification with created/skipped/error counts + deep link. Status surfaces in the existing notification bell.
- Persist in-app background run state (analogous to `WorkflowJobRunDb`) so progress survives reload.
- Threshold is a setting (reuse the configurable-limit plumbing) with a sensible default.

## Additional Context
Establishes the dual sync/async model reused by the domain-sync issue — build this first. No websocket/SSE (60s polling is the surface). Tests: threshold boundary selects sync vs background; background run emits progress then terminal notification with correct counts; sync path unchanged below threshold; prefer patching the executor boundary over real threads/Databricks.

larsgeorge-db:
## What & why

On large Unity Catalog estates the Schema Importer's synchronous `/import` blocks the request/UI for minutes with no progress and no safe way to walk away. This adds an **in-app background execution path** so large imports run off the request thread, report live progress, and notify on completion.

It mirrors the codebase's existing async precedent — the **Ontology Generator** (`OntologyGeneratorManager`): a DB-persisted run row + daemon thread on a fresh session + status polling. No Databricks Job is involved (that's a deferred follow-up, see below).

Part of the large-account scale batch (epic #763). PRD: `docs/prds/prd-large-account-scale-and-uc-sync.md`.

## Backend

- **`schema_import_runs`** table + repository + Alembic migration `m2_schema_import_runs` (descends from the current head `l1_entity_domain_associations`).
- **`SchemaImportRunsManager`** (app.state singleton): `start_run()` persists a run, creates a `JOB_PROGRESS` notification, and spawns a daemon thread that runs the import on a fresh session with a `threading.Event` cancel and a per-user concurrency cap. On completion it updates the same notification to a terminal **success / failed / cancelled** message with counts — so completion is reported in the notification area even if the user navigated away.
- **`execute_import()`** gains optional `progress_callback` + `cancel_event` (fully backward compatible; the synchronous path is unchanged). Progress persistence is throttled to ~1/s.
- **Routes**: `POST /import-async` (202 + `run_id`), `GET /runs`, `GET /runs/{id}` (poll target, `db.refresh`), `POST /runs/{id}/cancel`.
- **`SCHEMA_IMPORT_ASYNC_THRESHOLD`** setting (config default 200 + General Settings + env) gates when the UI offers a background run.

## Frontend

- The import preview dialog offers **"Run in background"** at/above the threshold, shows **live progress** (polls the run every 2s: processed/total + created/skipped/errors), supports **Cancel**, and lets the user **close the dialog while the import keeps running** (completion lands in the notification bell).

## Tests

- Repository CRUD + progress/status/result/concurrency-count.
- `SchemaImportRunsManager` start_run bookkeeping (run persisted, notification created + id stored, cancel event registered — thread patched out), concurrency-cap `ValueError`, and cancel semantics.
- Existing `execute_import`/browse tests still green (signature change is additive).

## Follow-ups (filed)

- **#827** — optional Databricks Job execution backend (deferred per review: "in-app now, Databricks Job as follow-up").
- **#826** — make *all* async/background processes report completion via the notification system (this PR is the reference implementation).

## Notes

- Migration descends from `l1_entity_domain_associations`, the current single head. #817 also descends from `l1`; when both land, the second will need a trivial merge migration — standard here.
- Background work uses the caller's forwarded OBO token (preserving their UC permissions), falling back to the app service principal when none is forwarded.

Closes #758


mvkonchits-db (approved):
✅ Approved — all CI passing.

mvkonchits-db:
Heads-up on the alembic chain (context: #861 collapsed the dual heads onto a single linear chain, and #833 adds `m1_assigned_users` as the new tail).

Once #833 merges, the tail of the migration chain will be:
`l1_entity_domain_associations -> m1_mcp_keyless_default -> m1_uc_domain_id -> m1_assigned_users`

This PR's `m2_add_schema_import_runs.py` currently roots at `l1_entity_domain_associations`, which will make it a second head. It needs to re-root onto the tail:

```python
down_revision: Union[str, Sequence[str], None] = "m1_assigned_users"   # was "l1_entity_domain_associations"
```
(and update the docstring `Revises:` line to match).

Separately, this branch is currently **CONFLICTING** with `development` in 7 files beyond the migration — `src/backend/src/common/config.py`, `controller/settings_manager.py`, `controller/schema_import_manager.py`, `models/schema_import.py`, and the three frontend files (`schema-importer/import-preview-dialog.tsx`, `types/schema-import.ts`, `views/schema-importer.tsx`). Leaving the rebase + conflict resolution to you since it touches the feature's own logic; happy to help re-verify `alembic heads` once it's rebased.

mvkonchits-db (approved):
Reviewed the full diff (multi-hat: correctness, concurrency, authz). Migration is idempotent (`_table_exists` guard), correctly re-rooted at `m1_assigned_users`, and `alembic heads` = single head `m2_schema_import_runs` (verified offline). Full CI suite green. Ownership scoping via `get_run_for_user`/`list_runs_for_user` is consistent — no IDOR. Faithfully mirrors the OntologyGeneratorManager in-app-thread pattern.

Approving. A few **non-blocking** follow-ups worth tracking (most are inherent to the in-app-thread + OBO design, which the PR already flags as interim pending the Databricks-Job backend):

1. **OBO token expiry on long imports (medium).** `_build_workspace_client` captures the caller's forwarded `x-forwarded-access-token` at request time. OBO tokens are ~1h; a large-estate async import (the feature's raison d'être) can outlive it, so connector/UC calls start 401-ing mid-run. The sync path never hits this (completes in-request). Consider SP fallback on token-expiry mid-run, or documenting the practical ceiling.
2. **Orphaned runs on restart (low-medium).** `daemon=True` threads die on deploy/restart, leaving the run row stuck `running` with no reconciler to mark it `failed`. A startup sweep (running→failed for rows with no live thread) would close this.
3. **Cancel is process-local (low).** `cancel_run` only sets the event if the thread lives in this replica's `_cancel_events`; on a multi-replica deployment a cancel landing on another replica silently no-ops, and the route ignores `cancel_run`'s return value (always 200). Fine for single-instance; worth a note if Ontos ever scales out.
4. **Concurrency-cap TOCTOU (low).** `count_running_for_user` → `create` isn't atomic, so two simultaneous starts can exceed `MAX_CONCURRENT_RUNS_PER_USER`. Soft cap, so acceptable.

None block merge.

Issue #688, opened 2026-08-07
[Bug] Approved access requests never issue a Unity Catalog grant — approve handler records "Active" status but is decoupled from the grant-delivery subsystem (related to #311)


surojitchowdhury:
### In simple terms

When a Data Owner clicks **Approve** on an access request, Ontos records the approval in its own database and tells the requester "Access granted — Active" — but on a normal workspace **no actual Unity Catalog grant is ever issued**. The user is shown a green "Active" badge and a success toast, yet never actually receives the privilege in Unity Catalog. So access *looks* granted in the app while the person still can't touch the data.

Ontos *does* contain real UC-grant code (it can call `WorkspaceClient.grants.update`), but that code lives in a separate "delivery" subsystem that the Approve button never calls. The grant only fires if an admin has separately configured an approval **workflow** with a delivery/grant step **and** turned on direct-delivery mode (which is **off by default**). The everyday Approve button bypasses all of that.

This is related to #311 ("Approved role access requests do not actually grant the role to the user"), but the root cause here is distinct and deeper: it is not a duplicate request store or a stale approval handler — it is **one approve handler that only persists status, decoupled from a grant-delivery subsystem that is never wired into it.**

### Current Behavior

Approving an access request via the Access panel persists app-internal records and notifies the requester, but issues no Unity Catalog grant. Verified end-to-end at commit `af801c98`:

**Single request store (no divergent second store):** `src/backend/src/db_models/access_grants.py`
- `:68` `access_grant_requests` (the pending request)
- `:124` `access_grants` (the app-internal "granted" record)

**Approve route:** `src/backend/src/routes/access_grants_routes.py:139` (`POST /api/access-grants/handle`) calls `manager.handle_request(...)` (`:164`) and returns `"message": "Request approved"` with a `grant` object (`:174`).

**Approve manager — DB + notification only, no UC:** `src/backend/src/controller/access_grants_manager.py:286` `handle_request`. On approve it does exactly three things and commits:
- `:318` `self._grant_repo.create(...)` — inserts an `access_grants` row
- `:331` `self._request_repo.update_status(..., APPROVED)` — marks the request approved
- `:344` `self._notify_requester_decision(..., approved=True)` — sends a notification
- `:372` `db.commit()`

[...]

PR #832, opened 2026-09-20
feat(domains): sync data domains with Unity Catalog (import + export) (#761)

larsgeorge-db:
## What & why

Implements **Sync Domains** (#761): bidirectional sync of data domains between Ontos and Unity Catalog.

The issue was written assuming UC domains are flat (no nesting) and proposed a `data_domain_N` tag hack to fake hierarchy. That premise is **outdated** — the Databricks SDK (0.137) now exposes a native `w.domains` `DomainsAPI` with real subdomains via `parent_domain_id`, **capped at exactly two levels** (root domains + one subdomain layer). This PR uses that native API directly.

> The `data_domain` / `data_domain_N` securable-tagging convention (`DomainExportAdapter`, `uc_tag_sync`) tags *securables* with domain membership — a different concern — and is left unchanged.

## Approach (confirmed with stakeholder)

- **Import (UC → Ontos):** read UC's domain list, rebuild its 2-level tree, and create/link Ontos domains — optionally under a chosen target root. Preserves hierarchy.
- **Export (Ontos → UC):** from a **configurable anchor**, the anchor's children map to UC root domains and its grandchildren to UC subdomains. Domains deeper than UC's two-level cap are **skipped and reported in the preview** (so an "Organization"-style root can be skipped, mapping level-2 → UC level-1).
- **Idempotent** by a new `uc_domain_id` column on `data_domains` (matched first, then by name).
- **Synchronous** with a dry-run preview that writes nothing (domains are low-cardinality). Async Databricks-Job execution is deferred (can later mirror the #758 runs pattern; ties into #826).

## Changes

**Backend**
- `common/uc_domains_client.py` — thin, testable wrapper over `w.domains` (list/create/update) + `UcDomain` dataclass; `PermissionDenied` → `DomainPermissionError` ("needs MANAGE DISCOVERY"); `tag_key` defaults to a slug of the name.
- `controller/domain_sync_manager.py` — `preview_*` / `execute_*` for both directions (level mapping, skip-too-deep, idempotency).
- `routes/domain_sync_routes.py` — `/api/domain-sync/{import,export}/{preview,execute}`; OBO workspace client, `data-domains` permission gate, audit logging.
- `data_domains`: new nullable indexed `uc_domain_id` column (migration `m1_uc_domain_id`) + read-model field.
- 10 unit tests: level mapping, skip-too-deep, idempotency by `uc_domain_id`, name-match backfill, and no-write previews.

**Frontend**
[...]

PR #833, opened 2026-09-20
fix(roles): grant role on access-request approval via individual user assignment (#760, #311)

### LINKED ISSUE #196: [Feature]: assigned_users data model + API round-trip
## Parent PRD

#195

## What to build

Add the `assigned_users` field (list of user email strings) to the App Role data model, persisted as a JSON text column identical to `assigned_groups`. This slice covers the full persistence round-trip: Alembic migration, SQLAlchemy column, Pydantic models, repository serialization/deserialization, and settings manager mapping — so that a role can be created/updated with `assigned_users` via the API and the field persists and returns correctly.

No authorization logic changes in this slice (that's a separate issue). No frontend changes.

### Key decisions from the PRD

- Column: `Text`, not null, default `'[]'` — same pattern as `assigned_groups`
- Migration revises `d1_odcs_v310_full_persistence` (current head)
- `AppRoleBase` gets `assigned_users: List[str]`, `AppRoleUpdate` gets `Optional[List[str]]`
- Repository serializes/deserializes with `json.dumps`/`json.loads`, same as `assigned_groups`
- Settings manager `_map_db_to_api()` deserializes and passes to `AppRole()` constructor
- Default role seeding includes `assigned_users: []`

## Acceptance criteria

- [ ] Alembic migration adds `assigned_users` column to `app_roles` table (Text, not null, default `'[]'`)
- [ ] `AppRoleDb` model has `assigned_users` column
- [ ] `AppRoleBase` and `AppRoleUpdate` Pydantic models include `assigned_users`
- [ ] Repository `create()` and `update()` serialize `assigned_users` to JSON
- [ ] Settings manager `_map_db_to_api()` deserializes `assigned_users` from JSON
- [ ] `GET /api/settings/roles` returns `assigned_users` array for each role
- [ ] `POST /api/settings/roles` and `PUT /api/settings/roles/{id}` accept and persist `assigned_users`
- [ ] Existing roles with no `assigned_users` return `[]` (backward compatible)

## Blocked by

None - can start immediately

## User stories addressed

- User story 1 (assign role directly by email — data layer)
- User story 7 (same permissions regardless of assignment method — data layer)
- User story 9 (existing group-based assignments unaffected)

### LINKED ISSUE #311: [Bug]: Approved role access requests do not actually grant the role to the user

### Current Behavior

Related: feature work in #195 / #196 / #197 / #198 will eliminate this gap. This issue tracks the user-facing symptom so it can be fixed (or labeled as a known    
  limitation) independently of the timing of #195.

When a user clicks "Request access" for an app role (e.g. Data Consumer):

  1. A workflow execution is triggered (`ON_REQUEST_ACCESS`) and an approval notification is sent to admins.
  2. The admin approves the request successfully.
  3. The requester receives a "Role Request Approved" notification.
  4. **However, the user is not actually given the role.** They cannot browse the app as a Data Consumer, and they can immediately submit the same request again.

  The approval handler `SettingsManager.handle_role_request_decision` (`src/backend/src/controller/settings_manager.py:2018-2149`) only logs the decision and creates
  notifications. The code explicitly comments that group assignment is expected to be handled by an external ITSM process:

  ```python
  # 3. Log decision (admin feedback only, no actual group assignment)
  if request_data.approved:
      logger.info(
          f"Role request APPROVED for {request_data.requester_email} (Role: {role_name}). "
          f"(Actual group assignment should be handled via external ITSM process)."
      )

  The default ON_REQUEST_ACCESS workflow ends in a pass step, which is a no-op (src/backend/src/common/workflow_executor.py:818-822). No step in the chain writes any
  record that would let the auth check resolve the user to the requested role.

  There is also no user_role_assignments table in the schema (src/backend/src/db_models/settings.py); AppRoleDb.assigned_groups only stores Databricks group names, so
  even a manual write has no clean target today.

### Expected Behavior

After an admin approves a role access request, the requester should be granted the role:

  - The role should appear in the user's permissions on next login (or refresh).
  - The user should not be able to re-request the same role while the assignment is active.
  - An audit record of who granted what when should be retrievable.

  If the deployment relies on an external ITSM process for group assignment, that integration should be a configurable option on the workflow, not the silent default.

### Steps To Reproduce

1. Deploy Ontos and log in as an Admin user.
  2. Settings → Roles: ensure a non-admin role (e.g. "Data Consumer") exists with at least one approver role and no group containing the test user in assigned_groups.
  3. Log in as a non-admin test user who currently has no role.
  4. Open the role access dialog and click Request access for "Data Consumer".
  5. Log back in as Admin and approve the request from the notification panel.
  6. Log in as the test user again.
  7. Observe:
    - The requester receives a "Role Request Approved" notification.
    - The user still has no role — the marketplace/Data Consumer features are not visible.
    - Clicking Request access for the same role is allowed again (no "already approved" state).

### Cloud

Azure

### Browser

Chrome

### Version

latest

### Relevant log output

```shell

```

### LINKED ISSUE #760: Approving a role access request does not grant the role

**Blocked by:** individual-user-to-role assignment (`assigned_users`) from `docs/prds/prd-individual-user-role-assignment.md` — not yet implemented.

## Current Behavior
A user without a matching app role requests access (`POST /api/user/request-role/{role_id}`), an approver approves (`POST /api/settings/roles/handle-request` → `handle_role_request_decision` in `settings_manager.py` ~L2278). Approval only logs, change-logs, and notifies — it never assigns the user to the role. The code comment (~L2345) says assignment "should be handled via external ITSM process". The requester gets an "approved" notification but still cannot use the feature.

## Expected Behavior
Approving assigns the requester to the target role so they immediately gain access. Denial changes nothing. The grant is recorded in the change log (who approved, who was granted).

## Steps To Reproduce
1. As a user with no matching app role, request access to a role.
2. As an approver, approve the request.
3. Observe the requester still lacks the role's permissions.

## Proposed fix / Additional Context
Depends on an individual-user-to-role assignment mechanism (`assigned_users` on `app_roles`), specified in `prd-individual-user-role-assignment.md` but **not yet implemented** (no `assigned_users` references exist in the codebase). Two orderings:
- **Preferred:** land individual-user assignment first, then on approve add the requester's email to the role's `assigned_users` and persist.
- **Fallback:** deliver the minimal `assigned_users` write path here (column + read-side matching + approval write), scoped down from that PRD's full UX.

Replace the "no actual group assignment" log with the assignment write; keep change-log (extended to record the grant), requester notification, and admin-notification-handled marking. Tests: approving adds requester to `assigned_users` and a previously-unauthorized user then passes the permission check; denial leaves assignments unchanged.

larsgeorge-db:
## What & why

Approving an app-role access request now **actually grants the role**. Previously `SettingsManager.handle_role_request_decision` only logged and notified — the step-3 comment literally said assignment "should be handled via external ITSM process" — so the requester was told "approved" but stayed locked out and could re-request forever.

**#311 is the same bug** (older report, same handler, same root cause); this closes both.

The fix requires a per-user grant mechanism (the app can't modify Databricks groups, and there's no persisted role-request table to read as membership), so this also lands the **individual-user assignment** feature (#196) it depends on, ported onto `development` from the stale `mvk/feat/issue_196_assigned_users_data_model` branch.

## Changes

**Backend — `assigned_users` (from #196)**
- `app_roles.assigned_users`: JSON-text email list (migration `m1_assigned_users`), parallel to `assigned_groups`; models/repo serialize + deserialize.
- `AuthorizationManager.get_user_effective_permissions` / `get_user_effective_role_ids`: a role matches when the user's groups intersect `assigned_groups` **OR** their email is in `assigned_users` (case-insensitive OR).
- `PermissionChecker` / `enforce_feature_permission` thread the caller email and **relax the no-groups guard** so an email-only assignment still resolves; `/api/user/(actual-)permissions` do the same.
- Requestable-roles resolution treats email-held roles as held, so an approved user can't re-request.

**Backend — the #760 fix**
- On approval, `handle_role_request_decision` appends the requester's email to the target role's `assigned_users` and records the grant in the change log. Denial changes nothing.

**Frontend (reuses the existing IdP `PrincipalPicker`, which emits UPN/email for users — consistent with Owners/Domains)**
- Role form: a user-mode `PrincipalPicker` bound to `assigned_users`.
- Roles table: a combined **"Principals"** column (groups → Users icon/dark badge; users → User icon/light badge; `+N more`).
- `checkUserHasRole` recognizes email assignment.

## Testing
- `test_role_email_assignment` (OR-matching, effective role-ids) and `test_role_approval_grant` (approval grants + user then resolves to the role's permissions; idempotent; denial is a no-op) — **10 passed**.
- Existing auth/settings/roles suites green with the additive signature changes (125 passed).
- `tsc --noEmit` clean. Single Alembic head (`m1_assigned_users`); column applies via the startup `alembic upgrade head` path.

> ⚠️ Note: `test_comments_manager.py` (8) and `test_settings_manager::test_get_settings_returns_dict` fail on **`development` already** (Comment model validation / global-comment permission / a missing `WORKSPACE_DEPLOYMENT_PATH` mock attr) — verified pre-existing with these changes stashed. Unrelated to this PR.

## Out of scope
- **#688** (access-grants → real Unity Catalog GRANT delivery) is a separate, deeper subsystem and stays open.

Closes #760
Closes #311
Closes #196


mvkonchits-db:
## Migration Sequencing Issue

⚠️ **Alembic single-head violation**: PR 833 migration `m1_assigned_users` descends from `l1_entity_domain_associations`, but so do:
- PR 817: `m1_mcp_keyless_default` → `l1`
- PR 828: `m2_schema_import_runs` → `l1`
- PR 832: `m1_uc_domain_id` → `l1`

When all merge, you'll have **4 separate heads** → alembic check will fail (requires single linear head).

### Solution: Merge sequentially

The PRs must merge in order, each rebasing the next:

1. ✅ Merge PR 817 → `m1_mcp_keyless_default` becomes new head
2. Rebase PR 833 migration to: `down_revision = 'm1_mcp_keyless_default'`
3. Merge PR 833 → `m1_assigned_users` becomes new head
4. Rebase PR 828 migration to: `down_revision = 'm1_assigned_users'`
5. Merge PR 828 → `m2_schema_import_runs` becomes new head
6. Rebase PR 832 migration to: `down_revision = 'm2_schema_import_runs'`
7. Merge PR 832 → `m1_uc_domain_id` becomes new head

This creates a linear migration chain and keeps alembic happy.

**Current queue status:**
- ✅ PR 816, 822, 830 queued (no migrations)
- 🔴 PR 833 needs rebase after 817 merges
- ⏸️ PR 828, 832 on hold (will need rebase after 833, then 828)

mvkonchits-db:
## 🚨 Code Review Findings (Deep Review)

PR 833 has **5 issues** that need to be addressed before merge:

### **1. RACE CONDITION: Concurrent role approvals lose updates** 🔴

**File:** `src/backend/src/controller/settings_manager.py:390`

**Failure scenario:** Two concurrent approvals to the same role:
- T1 reads `assigned_users=['user1@']`
- T2 reads `assigned_users=['user1@']`
- T1 appends user2@, commits
- T2 appends user3@, commits (overwrites T1's change)
- **Result:** user2@ is lost

The idempotent check only guards duplicate same-user approvals, not parallel different-user updates.

**Fix needed:** Use atomic operation or database-level locking on `assigned_users` updates.

---

### **2. Inconsistent null-handling crashes authorization_manager** ⚠️

**Files:** `authorization_manager.py:197` vs `settings_manager.py:327`

- **authorization_manager line 197:** `set(u.lower() for u in ...)` crashes on None values  
- **settings_manager line 327:** `(u or '')` handles gracefully

If `assigned_users` contains None → authorization_manager crashes while settings_manager doesn't.

**Fix needed:** Consistent defensive handling across both files.

---

### **3. Email normalization inconsistency** ⚠️

**File:** `settings_manager.py:390`

```python
# Line 390 (differs from lines 175/315):
email_norm = (request_data.requester_email or "").strip().lower()

# Lines 175/315 use:
(... or None)  # tail instead of ""
```

If email is whitespace-only: line 390 produces `""`, lines 175/315 produce `None`. Future code treating them differently could break.

**Fix needed:** Consistent pattern across all three locations.

---

### **4. Defensive list comprehension inconsistency** ⚠️

**Files:** `authorization_manager.py:255` vs `settings_manager.py:327`

- Line 255 lacks the `(u or '')` check present at 327
- If `assigned_users` contains non-string values → auth_manager crashes, settings_manager silently converts

**Fix needed:** Match defensive patterns.

---

### **5. Group name null coalescing hides data quality issues** ⚠️

**File:** `settings_manager.py:314`

```python
set((g or '').lower() for g in ...)
```

Converts None group names to empty strings, silently accepting malformed data instead of failing fast.

**Fix needed:** Either validate/reject malformed data, or document why silent conversion is acceptable.

---

## Summary

**Critical blocker:** Issue #1 (race condition) introduces a new concurrency bug in code meant to fix bug #760. All 5 issues stem from inconsistent defensive coding patterns.

**Recommendation:** Address race condition + normalize all null-handling patterns before merge. The fix is solid in concept but needs hardening for concurrent access and edge cases.

larsgeorge-db:
Thanks for the thorough review, @mvkonchits-db — all points addressed. Details below.

## Migration sequencing

Resolved, but slightly differently from the linear plan since the ground shifted while this sat in the queue: **#832 (`m1_uc_domain_id`) has already merged** to `development`, alongside `m1_mcp_keyless_default`. So `development` was briefly two-headed off `l1` on its own.

Rather than chain after `m1_mcp_keyless_default` only (which would leave `m1_uc_domain_id` as a second head), `m1_assigned_users` now **merge-points both** existing heads:

```
l1_entity_domain_associations → { m1_uc_domain_id, m1_mcp_keyless_default } → m1_assigned_users (mergepoint)
```

`alembic heads` reports a single head and the Alembic Single-Head check is green. This also repairs `development`'s own two-head state once this merges. (#828's `m2_schema_import_runs` is still separate and will rebase onto the head after this lands.)

## Code review findings

All five fixed in `ea4829f5`:

1. **Race condition** — `handle_role_request_decision` now takes a row lock (`SELECT … FOR UPDATE`) on the `app_roles` row around the `assigned_users` read-modify-write, so concurrent approvals to the same role serialize instead of clobbering each other (row lock on Postgres; harmless no-op on SQLite). Added `test_two_different_users_both_granted` to cover accumulation.
2 & 4. **Null-handling** — both `get_user_effective_permissions` and `get_user_effective_role_ids` now use `(x or '').lower()` on the group/user comprehensions, matching `settings_manager`, so a `None` entry can't crash `.lower()`.
3. **Email normalization** — the grant path now uses the `… or None` tail, consistent with the other `assigned_users` sites.
5. **Group null-coalescing** — kept the defensive `(g or '')` but documented why: this is an authorization read path, and failing fast on one malformed row would lock the user out of the whole app; empty strings never match a real group/email, so dirty data is simply ignored.

Branch is up to date with `development` and green on CI. Thanks again!

mvkonchits-db (approved):
Re-pointed onto the single-head chain per #861: dropped the redundant merge tuple, down_revision = "m1_uc_domain_id" (now a descendant of m1_mcp_keyless_default), updated docstring to drop merge-migration framing, merged latest development in. Verified offline `alembic heads` -> single head m1_assigned_users. All CI green.

PR #846, opened 2026-09-21
ci(docs): install npm deps via JFrog on protected runners

larsgeorge-db:
## What

The deploy now runs on the protected runner group (previous fix), but the **build** fails ([run](https://github.com/databrickslabs/ontos/actions/runs/35617901226)):

```
npm error Exit handler never called!
...
sh: 1: docusaurus: not found
Process completed with exit code 127
```

**Root cause:** on the `databrickslabs-protected-runner-group` runners, npm installs through the internal **JFrog registry** (authenticated via OIDC), not public npm. My workflow used stock `actions/setup-node`, which leaves npm pointed at a registry it can't reach from these runners, so `npm ci` aborts and `node_modules` is left incomplete — the `docusaurus` binary is never installed, hence exit 127 on `npm run build`.

## Fix

Use the repo's existing composite action `./.github/actions/setup-node-jfrog` — exactly what `test-coverage.yml` does for its frontend `npm ci` — pointing `lock-path` at the docs lockfile:

```yaml
- name: Checkout
  uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
- name: Setup Node.js (JFrog npm)
  uses: ./.github/actions/setup-node-jfrog
  with:
    node-version: '24'
    lock-path: website/ontos/package-lock.json
```

That fetches a JFrog OIDC token (the job already has `id-token: write`), writes `~/.npmrc`, and sets up Node with caching. Also bumps checkout to the v7 SHA used elsewhere (clears the Node 20 deprecation warning) and Node 24 to match the other jobs.

This is the last expected blocker before the site publishes at https://databrickslabs.github.io/ontos/.


mvkonchits-db (approved):
✅ Approved — CI-only fix, well-reasoned. Alembic failure is pre-existing (migration branching issue, unrelated to this PR).

PR #847, opened 2026-09-21
fix(docs): import Tailwind utilities so landing page is styled

larsgeorge-db:
## What

The site deploys and is live at https://databrickslabs.github.io/ontos/ 🎉 — but the landing page renders **unstyled** (logo + plain text, no layout).

**Root cause — not a deploy issue.** The deployed CSS is byte-identical to a local build; the site has simply always built *without Tailwind utilities*:

- The landing page (`src/pages/index.tsx`) uses Tailwind classes (`flex flex-col items-center w-full`, etc.).
- `docusaurus.config.ts` registers the `@tailwindcss/postcss` plugin.
- **But no CSS ever imported Tailwind.** In Tailwind v4, utilities are only emitted when the entry CSS imports them (`git log -S` confirms `custom.css` never had a Tailwind directive). So every Tailwind class resolved to nothing, and only Infima (Docusaurus base) shipped.

Doc pages looked fine because they rely on Infima + CSS modules; only the utility-class-driven landing page broke.

## Fix

Add to `src/css/custom.css`:

```css
@import 'tailwindcss/theme.css' layer(theme);
@import 'tailwindcss/utilities.css' layer(utilities);
```

Importing the **theme + utilities layers only** (not the full `tailwindcss` bundle) is deliberate: the full import also pulls in **Preflight**, whose base-element reset conflicts with Infima and would break Docusaurus doc-page styling.

## Verified locally

- `npm run build` succeeds; output CSS hash changed `a9faf928` → `46593d55`.
- Tailwind utilities now emitted: `.flex-col`, `.items-center`, `.w-full` (v4 tree-shakes to only-used classes).
- **No Preflight universal reset** in the output CSS — Infima doc styling stays intact.


mvkonchits-db (approved):
✅ Approved — clean CSS fix. Tailwind utilities properly scoped to avoid Preflight conflict with Infima.

PR #848, opened 2026-09-21
fix(docs): vendor full Infima light :root block so site renders styled

larsgeorge-db:
## What

The deployed site (https://databrickslabs.github.io/ontos/) rendered **completely unstyled** — browser serif default, no layout, plain-text buttons.

**Root cause (confirmed live in-browser via injected CSS + local-build screenshots):** Docusaurus 3.10's default webpack bundler minifies CSS with **cssnano**, which **drops Infima's entire light-mode `:root { ... }` variable block**. Every base `--ifm-*` token (fonts, sizes, spacing, colors, navbar, buttons) vanishes from the output; the `[data-theme='dark']` block survives. With the light-mode base vars gone, `html { font: ... var(--ifm-font-family-base) }` resolves empty → invalid shorthand → the whole theme collapses to serif/unstyled.

On the live page I measured: `--ifm-font-family-base`, `--ifm-font-size-base`, `--ifm-navbar-height`, `--ifm-background-color`, etc. all **EMPTY**; only `--ifm-color-primary` resolved (because we define it in `custom.css`).

## Fix

Variables authored in `custom.css` **do** survive the build (verified). So this vendors **Infima's complete light `:root` block** (infima 0.2.0-alpha.45, copied verbatim) into `custom.css`, with our brand-color overrides kept in a separate `:root` after it so they still win.

Verified via local `npm run build` + served screenshot: fonts, outlined buttons, search box, navbar GitHub icon, dark footer, and brand-blue links all render correctly. All base `--ifm-*` vars present in the built CSS.

> Note: this branch's earlier commits were a font-only attempt that proved insufficient (2 of ~284 stripped vars). The final commit supersedes them.

## Follow-up

This is a **workaround** for the cssnano bug. Tracking issue **#849** covers switching the build to `@docusaurus/faster` (LightningCSS minifier, which doesn't strip the block) and deleting the vendored block — currently blocked because `@docusaurus/faster`'s transitive dep isn't on our JFrog npm proxy. A `TODO(#849)` marks the block in `custom.css`.


danielrozo-db (approved):
Workaround to fix Docusaurus gitpages deployment.

